PR #3, opened 2016-09-27
Vendor!

malnick:
@dcos/day2ops Any objections to this?


rji:
Could we use something like [godep](https://github.com/tools/godep) instead, so that we don't commit a bunch of vendored code to the repo?

rji:
Erm, ignore me, looks like I'm just behind the times :)

rji:
The only thing I'm thinking about here is if we can move the `vendor/` directory inside the `collector/` directory, since this one repo contains several different components to the overarching "dcos-metrics" project (collector, consumers, Mesos module, and so on).

malnick:
Can we do that in a separate commit? Right now, that could damage how we import the project and the GOPATH. For example, the common function here is to have a lib in the root of the repo, with a subdir for `cmd` which implements a command (executable) around that lib. This lib is definitely going to change per the work you're doing, so I'd rather leave the reworking of the repo structure up to you at that point in time. 

You can also pull out the module to another repo in another commit too at that point.

rji:
👍

PR #6, opened 2016-09-29
Unify build/test scripts and gitignore files

rji:
Prior to this PR, dcos-metrics relied on several Makefiles and shell scripts to build and test the components of this project. Beginning with this commit, we attempt to unify these build scripts into a DRY methodology.

This PR begins with the `collector/` component, but these scripts will be expanded to cover the Mesos module and consumers at some point as well.

This PR also includes fixes based on recommendations from goimports and golint.


rji:
Ping @dcos/day2ops for review.

I have a Jenkins job set up at https://jenkins.mesosphere.com/service/jenkins/job/dcos-metrics/job/public-dcos-metrics-master/ that will use these scripts, and the build badge in the README tracks this job (master branch).

Ideally I'd like this to be a Jenkinsfile that we can commit to the repo, but when I was building the config earlier I found that some of the plugins we depend on (Cobertura, for example) don't have entries in the DSL yet (or at least not on *our* CI cluster) 😕 

I also need to add some docs on how to use these new scripts. I was planning on doing that on a later PR when I go back through and overhaul our docs, but I'm happy to add them here if anyone feels strongly about it.

.gitignore
```diff
@@ -1,6 +1,61 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
malnick: How did this list get so big?

Makefile
```diff
@@ -0,0 +1,10 @@
+all: clean build-collector test-collector
+
+build-collector:
```
malnick: can we use just `build`?

Makefile
```diff
@@ -0,0 +1,10 @@
...
+	bash -c "./scripts/build.sh collector"
+
+test-collector:
```
malnick: and just `test`?

scripts/build.sh
```diff
@@ -1,5 +1,22 @@
...
+
+rm -rf $BUILD_DIR && mkdir -p $BUILD_DIR
+cd "${SOURCE_DIR}/${COMPONENT}"
+
+# build metrics-schema package
+go generate
+
+# build binary
+cd cmd/dcos-metrics-collector
+go build -a -ldflags "-w" -o ${BUILD_DIR}/dcos-metrics-collector-${GIT_REF} *.go
```
malnick: Why omit DWARF symbols? https://golang.org/cmd/link/

malnick: :+1: for the naming convention.

scripts/test.sh
```diff
@@ -0,0 +1,129 @@
...
+BUILD_DIR="${SOURCE_DIR}/build/${COMPONENT}"
+
+
+function test_collector {
+    local test_suite="$1"
+    local test_dirs="collector/"
+    local package_dirs="./collector/..."
+    local ignore_packages=""
+
+    if [[ $test_suite == "unit" ]]; then
```
malnick: This. Is. Awesome.

malnick:
Also, when can we get this as a git hook to this repo to block our PRs :)

.gitignore
rji: It combines recommended .gitignore files for Go, C++, _and_ Java, instead of maintaining separate gitignore files in each component's directory. It's a combination of what was already in the repo, and recommendations from https://github.com/github/gitignore.

Makefile
rji: Sure! I wasn't sure if it would ever be desirable to build these individually, but yeah I think `build` and `test` will be simpler for now. I'll push up a commit that changes this.

Makefile
rji: Totally!

scripts/build.sh
rji: My thoughts were to omit DWARF symbols (`-w`) for production use because, more often than not, we _probably_ don't need to debug this service after it's been built. I figured if worse comes to worse and we're trying to debug something in a live environment that we can't get through traditional troubleshooting, we can release a `-debug` build for that specific case. If you know of a reason that we might want to keep these around, I'm happy to hear the counterpoint!

This was inspired by https://github.com/golang/go/issues/6853#issuecomment-66088650 (and other various blog posts on the internet).

scripts/test.sh
rji: Why thank you sir 😄

rji:
@malnick Apparently GitHub comments don't collapse anymore when I push a commit that addresses a comment. Bummer 😞. I responded / updated this PR based on your feedback.

I could probably create the PR job and run this PR through it now if you don't have any additional feedback for me on the build scripts. Just didn't want to do any repeat work while developing this. How does that sound to you?

scripts/build.sh
malnick: It's true that I've never had to use GDB or other debuggers on my built executables, but this may come in handy down the road. The trade off is a significantly larger binary (by percent, approx 25%) however, when we're talking about at most tens of megabytes, this doesn't really concern me, and I'd rather have an ability to debug until we have complete stability than not.

rji: Makes sense, I'll turn them back on

rji:

rji:

PR #7, opened 2016-09-29
Agent Configuration

malnick:
"The Big Plan"(TM) - break down agent_poller.go piece by piece, forming a complete Agent{} struct with methods on it that can be broken out into a generic `HostPoller interface{}` later so we can build a Master version. This first bit removes the library level flags, and instead passes them into the library through the `NewAgent() -> Agent{}` construct. 

- [x] Agent poller becomes a obejct
- [x] Agent poller sets config from executable code
- [x] Refactor code to enable master poller and generic role interface in the future

### DIFF
--- a/collector/agent_poller.go
+++ b/collector/agent_poller.go
@@ -15,8 +15,6 @@ import (
 )
 
 var (
-	agentIpCommandFlag = StringEnvFlag("agent-ip-command", "/opt/mesosphere/bin/detect_ip",
-		"A command to execute which writes the local Mesos Agent's IP to stdout")
 	agentPortFlag = IntEnvFlag("agent-port", 5051,
 		"HTTP port to use when querying the local Mesos Agent")
 	agentPollPeriodFlag = IntEnvFlag("agent-poll-period", 15,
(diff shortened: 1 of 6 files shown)


collector/agent_poller.go
```diff
@@ -69,80 +65,120 @@ type AgentState struct {
...
+	AgentStateChan chan<- *AgentState
+}
+
+func NewAgent(
+	ipCommand string,
+	port int,
+	pollPeriod int,
+	topic string) (Agent, error) {
+	a := Agent{}
+	if len(ipCommand) < 0 {
```
malnick: Error messages should use the YAML keys not the object keys

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
 	executorAppNames map[string]string
 }
 
+type Agent struct {
```
malnick: Should eventually include AgentState{} but in a later PR

malnick:
@dcos/day2ops This is ready for review.

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
+	AgentStateChan chan<- *AgentState
+}
+
+func NewAgent(
+	ipCommand string,
+	port int,
+	pollPeriod int,
+	topic string) (Agent, error) {
+	a := Agent{}
+	if len(ipCommand) < 0 {
```
darkonie: i don't think a string length can have a negative value

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
+		return a, errors.New("Must pass ipAddress to NewAgent()")
+	}
+	if port < 1024 {
+		return a, errors.New("Must pass port to NewAgent()")
+	}
+	if pollPeriod == 0 {
+		return a, errors.New("Must pass pollPeriod to NewAgent()")
+	}
+	if len(topic) < 1 {
```
darkonie: this also looks weird
if you want to check if the string is empty you can do
`if topic == ""`
or
`if len(topic) == 0`
i prefer first, because it explicitly tell that you are comparing a string not a slice

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
+	return a, nil
+}
+
 // Runs an Agent Poller which periodically produces data retrieved from a local Mesos Agent.
 // This function should be run as a gofunc.
-func RunAgentPoller(recordsChan chan<- *AvroDatum, agentStateChan chan<- *AgentState, stats chan<- StatsEvent) {
+func (a *Agent) Run(recordsChan chan<- *AvroDatum, stats chan<- StatsEvent) {
 	// fetch agent ip once. per DC/OS docs, changing a node IP is unsupported
-	var agentIp string = ""
+	a.AgentIP = ""
```
darkonie: unless you explicitly set `a.AgentIP` to empty string this is redundant. Strings are empty by default

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
-		agentIp = getAgentIp(stats)
+		//TODO needs err
+		a.getIP(stats)
 	}
 
 	// do one poll immediately upon starting, to ensure that agent metadata is populated early:
-	pollAgent(agentIp, recordsChan, agentStateChan, stats)
-	ticker := time.NewTicker(time.Second * time.Duration(*agentPollPeriodFlag))
+	a.pollAgent(recordsChan, stats)
+	ticker := time.NewTicker(time.Second * time.Duration(a.PollPeriod))
```
darkonie: rather then passing integer `a.PollPeriod`, can we pass a duration string and read is with `ParseDuration`
https://golang.org/pkg/time/#ParseDuration
you can take a look at this example:
https://blog.golang.org/context Look for this line `time.ParseDuration(req.FormValue("timeout"))`

this will give us much more flexibility

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
 	}
 }
 
 // runs detect_ip => "10.0.3.26\n"
-func getAgentIp(stats chan<- StatsEvent) string {
+func (a *Agent) getIP(stats chan<- StatsEvent) error {
 	stats <- MakeEvent(AgentIpLookup)
-	cmdWithArgs := strings.Split(*agentIpCommandFlag, " ")
+	cmdWithArgs := strings.Split(a.IPCommand, " ")
 	ipBytes, err := exec.Command(cmdWithArgs[0], cmdWithArgs[1:]...).Output()
```
darkonie: there is one really big issue here
suppose you run external command with `exec.Command` but the command never exists. Your entire program will be blocked forever. We should always assume we might timeout.
Take a look at https://golang.org/pkg/context/#WithTimeout

collector/agent_poller.go
```diff
@@ -383,12 +415,12 @@ func getAgentState(agentIp string, stats chan<- StatsEvent) (*AgentState, error)
...
 }
 
-func getJsonFromAgent(agentIp string, urlPath string, testFileFlag *string, stats chan<- StatsEvent) (*jason.Value, error) {
+func (a *Agent) getJsonFromAgent(urlPath string, testFileFlag *string, stats chan<- StatsEvent) (*jason.Value, error) {
 	stats <- MakeEvent(AgentQuery)
 	var rawJson []byte = nil
 	var err error = nil
 	if len(*testFileFlag) == 0 {
-		endpoint := fmt.Sprintf("http://%s:%d%s", agentIp, *agentPortFlag, urlPath)
+		endpoint := fmt.Sprintf("http://%s:%d%s", a.AgentIP, a.Port, urlPath)
```
darkonie: this is also not the right way of doing it, can we encode a URL with https://golang.org/pkg/net/url/#URL
there is an example

collector/cmd/dcos-metrics-collector/config.go
```diff
@@ -8,18 +8,35 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-type ConfigFile struct {
-	PollAgent     bool `yaml:"poll_agent"`
-	HttpProfiler  bool `yaml:"http_profiler"`
-	KafkaProducer bool `yaml:"kafka_producer"`
-	ConfigPath    string
+type MasterConfig struct {
```
darkonie: All exported structures need comments, linter will complain

collector/cmd/dcos-metrics-collector/config.go
```diff
@@ -32,26 +49,30 @@ func (c *ConfigFile) loadConfig() error {
...
 	// Override default config with CLI flags if any
 	if err := thisFlagSet.Parse(args); err != nil {
 		fmt.Println("Errors encountered parsing flags.")
 		return c, err
 	}
+
 	if err := c.loadConfig(); err != nil {
 		return c, err
 	}
+	fmt.Printf("%+v", c)
```
darkonie: this should not be here

collector/cmd/dcos-metrics-collector/config.go
rji: The linter is going to complain about a lot of things at the moment. I have a separate commit that fixes dozens of these cases, but we're waiting until after this PR is reviewed and merged until I apply it 😄

rji: Last count I think there were ~75 issues.

collector/agent_poller.go
darkonie: Also it doesn't look like we are validating the output. Can we validate the IP address?

collector/agent_poller.go
```diff
@@ -69,80 +61,120 @@ type AgentState struct {
...
+	a.AgentIP = ip
+
+	return nil
 }
 
 // fetches container-level resource metrics from the agent (via /containers), emits to the framework topics (default 'metrics-<framework_id>')
-func getContainerMetrics(agentIp string, agentState *AgentState, stats chan<- StatsEvent) ([]*AvroDatum, error) {
-	rootJson, err := getJsonFromAgent(agentIp, "/containers", agentTestContainersFileFlag, stats)
+func (a *Agent) getContainerMetrics(agentState *AgentState, stats chan<- StatsEvent) ([]*AvroDatum, error) {
+	rootJson, err := a.getJsonFromAgent("/containers", agentTestContainersFileFlag, stats)
```
darkonie: should be `a.getJSONfromAgent`, linter wouldn't like it

collector/cmd/dcos-metrics-collector/config_test.go
```diff
@@ -9,15 +9,15 @@ import (
 )
 
 func TestDefaultConfig(t *testing.T) {
-	testConfig := defaultConfig()
+	testConfig := newConfig()
 
 	testFileType := reflect.TypeOf(testConfig)
-	if testFileType.Name() != "ConfigFile" {
+	if testFileType.Name() != "CollectorConfig" {
```
darkonie: this looks really weird, why would we do that?
`newConfig()` will always return type `CollectorConfig`

collector/agent_poller.go
malnick: `topic` is an int.

collector/agent_poller.go
malnick: The poll period is used by a lot of other code in the library, so I wanted to minimize this refactor to some extent (no go hog wild with moving so much around, at the cost of making a very large PR). Can we add this to a TODO in a later PR?

collector/agent_poller.go
malnick: Yup, those are comments I address in the comment of the code. I was hoping to reuse the code you have in 3DT for a shared method that does what your code already implements. Right now we only have agent_poller, but I want to make this even more generic with a HostPoller interface and a HostUtiliities library that either pulls in your 3DT/shared code that does ip script execution or at minimum share between the implementations of a poller.

collector/agent_poller.go
malnick: Today, these test flags are scattered around the library. I did not want to remove them in this PR since i'm not 100% sure how to use them properly. This PR removes the main configuration flags. In a later iteration I want to move the test flags into the executable code configuration, and leverage URL to validate the flag options there.

collector/cmd/dcos-metrics-collector/config.go
malnick: Good catch!

collector/agent_poller.go
malnick: See @rji comments above. There is a commit to update all the `Ip` -> `IP` and the like in the pipeline.

collector/cmd/dcos-metrics-collector/config_test.go
malnick: I'm not sure why I put this here, given the promises of strict typing and all :)

collector/agent_poller.go
malnick: https://github.com/dcos/3dt/blob/master/api/helpers.go#L54

^^ I'm going to update it to use this, and remove the IP Command from the config. It's safe to bet this is available to our executable.

malnick:
@darkonie I addressed most of the comments. I'm leaving method refactoring out of this PR, I want to keep it scoped to getting the Agent{} struct in and a configuration set from the executable. I will address the follow up in the next PR.

malnick:

PR #10, opened 2016-10-01
Generic Consumer [WIP]

malnick:
## Overview
1. Add a generic consumer through a `Metric{}` struct and `Metricer interface{}` 
   1. Better nomenclature suggestions appreciated. `metricizer`? `metricator`?
   2. The idea is to have a single CLI interface that can consume metrics from the various sources out there. 
2. Add a `KafkaMetric{}` which implements the `Metricer interface{}`. 
   1. This new construct allows us to easily add a `InfluxMetric{}` or `KairosMetric{}` which also satisfy the `Metricer interface{}` in the future.
3. Add a `action` package which supports printing to STDOUT.
   1. We can add more actions to the package and thus to the CLI in the future. 
## CLI

Implemented with `urfave/cli` (formerly codegansta). 

Currently, it only supports one action `print`:

```
./consumer-c0308e5 -help
NAME:
   DC/OS Metrics Consumer: Consumer Metrics From Anywhere! - A new cli application

USAGE:
   consumer-c0308e5 [global options] command [command options] [arguments...]

VERSION:
   consumer-0.0.1 @ revision c0308e5

COMMANDS:
     print, k  Print metrics to stdout for a given consumer
     help, h   Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help, -h     show help
   --version, -v  print the version
```

And one consumer subcommand for print, `kafka`:

```
./consumer-c0308e5 print -help
NAME:
   DC/OS Metrics Consumer: Consumer Metrics From Anywhere! print - Print metrics to stdout for a given consumer

USAGE:
   DC/OS Metrics Consumer: Consumer Metrics From Anywhere! print command [command options] [arguments...]

COMMANDS:
     kafka  Use the kafka consumer

OPTIONS:
   --help, -h  show help
```

The `kafka` subcommand supports several flags for kafka-specific configuration:

```
./consumer-c0308e5 print kafka -help
NAME:
   DC/OS Metrics Consumer: Consumer Metrics From Anywhere! print kafka - Use the kafka consumer

USAGE:
   DC/OS Metrics Consumer: Consumer Metrics From Anywhere! print kafka [command options] [arguments...]

OPTIONS:
   --brokers value    Brokers <host:9092,host:9092,..> (default: "localhost:9092") [$CONSUMER_CONF_KAFKA_BROKERS]
   --topic value      Kafka topic to get [$CONSUMER_CONF_KAFKA_TOPIC]
[...]

PR #12, opened 2016-10-14
Refactor, part one

rji:
This is part one in refactoring the dcos-metrics project. It reorganizes a lot of the existing code and adds a number of new test cases. In summary:
* Vendor goconvey, logrus and relocate vendor directory to the root of the repo so that it can be used by additional go code in the future (i.e., #10).
* Adds copyright statements to each of the source files
* Consolidates `main` packages
* Relocates collector test scripts
* Flattens collector package
* Modifies test script for local development
* Apply a number of `gofmt`, `goimports`, and `golint` fixes

I've flattened the existing `collector` package in preparation of breaking it up into smaller standalone packages, such as splitting out the idea of `producers` from the `collector` code. But I'd like for this to land before I start splitting that up.


collector/statsdEmitter_test.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestToStatsdLabel(t *testing.T) {
+	Convey("When converting a StatsEvent to a StatsD label", t, func() {
+
+		// Spot check a few event types
+		Convey("Known events should produce valid labels", func() {
+			So(toStatsdLabel(
+				StatsEvent{evttype: 2}), // 2 => TCPResolveFailed
```
rji: Hey @nickbp, quick questionL when I was testing `toStatsdLabel` I noticed that the second entry in the `const()` starts the `iota`. Without explicitly setting this to `0`, it appears that these event types actually start with 2.

Since I was just adding tests for the existing code, I _assumed_ this was the correct behavior. Just wanted to ping you and get a 👍 or 👎 either way.

malnick: Let's not halt work due to this, we can come back and update the test once @nickbp is back in the office.

malnick:
:ship:

collector/statsdEmitter_test.go
rji: Sounds good, thanks!

nickbp: It shouldn't be a problem as only the enum labels are used; the number values themselves aren't written/sent anywhere. If possible I think it'd be ideal if this code checked the labels directly rather than their integer values, as the latter may be brittle as the entry list evolves.

rji: Oh cool, thanks @nickbp. I was originally checking the labels but thought the ints might be important. I'll fix that up in a separate commit.

PR #13, opened 2016-10-18
Malnick/breakout producers

malnick:
# This PR aims to:
- [x] Break out Kafka Producer
- [x] Break out Statsd Producer
- [x] Refactor to use config file over flags where possible
- [x] Update build script to use root `dcos-metrics-collector.go`

# What we ended up with
## Events Package
This is formerly the statsdEmitter stuff. It needs to be further refactored to move the statsd producer code into the the `producer` package.

## Utils (Name TBD)
This is utilities used by other libraries. Utils should never import a library from within the dcos-metrics project to ensure we don't have cyclical imports. Currently it has the HTTP utility functions formerly in `collector` package.

We should probably use a far less generic naming convention for this package, such as `MetricsUtil` or `DCOSMetricsUtil`. Suggestions appreciated.


rji:
Hey @malnick, I updated this branch/PR with the following:
* relocate statsd from `events` to `producers`
* relocate the generated `metrics_schema` package to the `schema/` directory, since it is used by more than just the collector
* removed (now unneeded) collector/generator.go (previously used for generating the go code from the Avro metrics schema)
* relocate the HTTP pprof runner into the `util` package
* minor build script tweaks

dcos-metrics-collector.go
```diff
@@ -39,7 +44,7 @@ type AgentConfig struct {
 // Config ...
 type Config struct {
 	HTTPProfiler  bool   `yaml:"http_profiler"`
-	KafkaProducer bool   `yaml:"kafka_producer"`
+	KafkaProducer bool   `yaml:"kafka_producer_enabled"`
```
rji: Would it be worthwhile to change this to a more generic Producer, and have the user specify one? As opposed to enabling / disabling the Kafka producer itself.

util/httpUtil.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	userAgent = "metrics-collector/1.0"
+	userAgent = "dcos-metrics-collector/1.0"
```
rji: TODO (at a future date): set the version (e.g., `1.0`) in `dcos-metrics-collector.go` and use that constant here, instead of burying the version somewhere in the `util` package.

dcos-metrics-collector.go
malnick: I like that idea, though we should probably do this in a later PR so we don't over reach what we refactor here.

malnick:
:ship: it!

PR #14, opened 2016-10-18
HTTP Producer [WIP]

malnick:
This PR contains the HTTP producer, and is based on #13


producers/interface.go
```diff
@@ -0,0 +1,43 @@
...
+// implement in order to receive, process, and present metrics to the caller or
+// client. All producers must use the MetricsMessage structure to receive
+// metrics, and they must implement their own struct for handling configuration.
+//
+// Further, although it isn't defined in this interface, it is recommended that
+// producers must also create their own MetricsMessage channel to be
+// used both in the implementation (e.g., &producerImpl{}) and to be returned
+// to the caller. Doing so ensures that, in the future, multiple producers
+// can be enabled at once (each producer has a dedicated chan).
+type ProducerInterface interface {
```
malnick: Small nit: can we call this a `MetricsProducer{}` since it's already an interface (doesn't need to be part of the name)?

producers/interface.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2016 Mesosphere, Inc.
```
malnick: :+1: on adding the (c) comments!

producers/interface.go
rji: Good suggestion, I like that!

Makefile
```diff
@@ -1,3 +1,5 @@
+.PHONY: all build test clean
```
malnick: does `test` and `all` need to be a phony target?

collector/agentPoller.go
```diff
@@ -115,39 +114,39 @@ func NewAgent(ipCommand string, port int, pollPeriod int, topic string) (Agent,
 
 // Run runs an Agent Poller which periodically produces data retrieved from a local Mesos Agent.
 // This function should be run as a gofunc.
-func (a *Agent) Run(recordsChan chan<- *AvroDatum, stats chan<- statsd.StatsEvent) {
+func (a *Agent) Run(recordsChan chan<- *AvroDatum) {
```
malnick: This isn't a MVP requirement, but it would be great to go straight to something like a `Datum{}` struct or `JSONDatum{}` struct instead of Avro. But I understand if that's not possible right now due to the Great Refactor (TM).

collector/agentPoller.go
```diff
@@ -429,8 +406,7 @@ func (a *Agent) getAgentState(stats chan<- statsd.StatsEvent) (*AgentState, erro
 		executorAppNames: executorAppNames}, nil
 }
 
-func (a *Agent) getJSONFromAgent(urlPath string, testFileFlag *string, stats chan<- statsd.StatsEvent) (*jason.Value, error) {
-	stats <- statsd.MakeEvent(statsd.AgentQuery)
+func (a *Agent) getJSONFromAgent(urlPath string, testFileFlag *string) (*jason.Value, error) {
 	var rawJSON []byte
```
malnick: Great opportunity to set use a literal struct to unmarshal JSON into, no need to use the `jason` lib for this, and we can be more explicit and clear about what data we're getting/need.

collector/agentPoller.go
```diff
@@ -429,8 +406,7 @@ func (a *Agent) getAgentState(stats chan<- statsd.StatsEvent) (*AgentState, erro
 		executorAppNames: executorAppNames}, nil
 }
 
-func (a *Agent) getJSONFromAgent(urlPath string, testFileFlag *string, stats chan<- statsd.StatsEvent) (*jason.Value, error) {
-	stats <- statsd.MakeEvent(statsd.AgentQuery)
+func (a *Agent) getJSONFromAgent(urlPath string, testFileFlag *string) (*jason.Value, error) {
```
malnick: Would love to rename this method to something like `queryMesosAPI`

collector/agentPoller.go
```diff
@@ -341,28 +327,25 @@ func (a *Agent) getSystemMetrics(agentState *AgentState, stats chan<- statsd.Sta
...
 	}
 
 	json, err := rootJSON.Object()
 	if err != nil {
-		stats <- statsd.MakeEvent(statsd.AgentQueryBadData)
 		return nil, err
 	}
 
 	// state["id"] (agent_id)
 	agentID, err := json.GetString("id")
```
malnick: Per comment below about using the `jason` lib: if we used literal structs to unmarshal our data we could simply set `agentID := agentData.AgentID` or something like that. It's far more clear IMO.

collector/agentPoller.go
```diff
@@ -341,28 +327,25 @@ func (a *Agent) getSystemMetrics(agentState *AgentState, stats chan<- statsd.Sta
...
 	if err != nil {
-		stats <- statsd.MakeEvent(statsd.AgentQueryBadData)
 		return nil, err
 	}
 
 	frameworks, err := json.GetObjectArray("frameworks")
```
malnick: Same here, and in the other instances of setting these values (per above comment).

collector/agentPoller.go
```diff
@@ -187,15 +183,14 @@ func (a *Agent) getIP(stats chan<- statsd.StatsEvent) error {
...
 // fetches container-level resource metrics from the agent (via /containers), emits to the framework topics (default 'metrics-<framework_id>')
-func (a *Agent) getContainerMetrics(agentState *AgentState, stats chan<- statsd.StatsEvent) ([]*AvroDatum, error) {
-	rootJSON, err := a.getJSONFromAgent("/containers", agentTestContainersFileFlag, stats)
+func (a *Agent) getContainerMetrics(agentState *AgentState) ([]*AvroDatum, error) {
+	rootJSON, err := a.getJSONFromAgent("/containers", agentTestContainersFileFlag)
 	if err != nil {
 		return nil, err
 	}
 
 	containersArray, err := rootJSON.ObjectArray()
```
malnick: Same comment here about using `jason` lib versus a literal struct that we can unmarshal to (see below comment in `getJSONFromAgent()`).

collector/agentPoller.go
rji: Already done, just not committed / pushed yet.

collector/agentPoller.go
rji: This method has been removed but I haven't committed and pushed yet.

collector/agentPoller.go
rji: Agreed, already done but haven't committed / pushed yet. It's ticketed internally though.

rji:
Existing tests are now green, but this will need more tests and a bit more work on the HTTP producer side before it's ready for review. Hang in there :)

rji:
The HTTP producer will need something like this going forward: https://github.com/dcos/dcos-go/pull/14

rji:
@malnick Despite CI failing, I think this is probably at a point where we can review it and fix up anything that is found during the review, including any failing tests / data races. I'd like to get this on master before context switching away from the HTTP producer and on to OS-level metrics.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
 
-// fetches container state from the agent (via /state) to populate AgentState
-func (a *Agent) getAgentState(stats chan<- statsd.StatsEvent) (*AgentState, error) {
-	rootJSON, err := a.getJSONFromAgent("/state", agentTestStateFileFlag, stats)
-	if err != nil {
-		return nil, err
+// transform will take metrics retrieved from the agent and perform any
+// transformations necessary to make the data fit the output expected by
+// producers.MetricsMessage.
+func (a *Agent) transform(in metricsMeta) (out []producers.MetricsMessage) {
```
rji: This method is a bit larger than I would normally care for, and there is some duplicated code between "produce agent metrics" and "produce container metrics." This can probably be extracted into a helper function that builds a new `producers.MetricsMessage{}` and returns, and might make testing a bit easier. :thinking:

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
-	frameworkNames := make(map[string]string, len(frameworks))
-
-	// state["frameworks"][N]["executors"][M]["id"] (executor_id)
-	// => state["frameworks"][N]["executors"][M]["labels"][L(MARATHON_APP_ID)]["value"] (application_name)
-	executorAppNames := make(map[string]string, 0)
+	return out
+}
 
+// getFrameworkInfoByFrameworkID returns the FrameworkInfo struct given its ID.
+// For example: "5349f49b-68b3-4638-aab2-fc4ec845f993-0000" => "marathon"
```
rji: Nit: this example no longer applies; this function was generalized to return a `frameworkInfo` instead of a `frameworkInfo.Name`.

collector/client.go
```diff
@@ -0,0 +1,117 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+var (
+	// TODO(roger): place the value of userAgent somewhere convenient
```
rji: Outstanding TODO item. Is there a way to programatically get the tag / SHA of the project when it's build  and inject it here? :thinking:

collector/client.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// NewHTTPClient returns a new instance of the HTTP client.
+func NewHTTPClient(host string, path string, timeout time.Duration) *HTTPClient {
+	var token string
+	// TODO(roger): if auth is required, get a token via jwt/transport. See
+	// https://github.com/dcos/dcos-go/blob/59a7af3/jwt/transport/cmd/example/main.go#L32
+	//
+	// This should be keyed off a configuration option.
+	token = ""
```
rji: Outstanding TODO.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+	NetRxBytes   uint64 `json:"net_rx_bytes,omitempty"`
+	NetRxErrors  uint64 `json:"net_rx_errors,omitempty"`
+	NetRxDropped uint64 `json:"net_rx_dropped,omitempty"`
+	NetTxPackets uint64 `json:"net_tx_packets,omitempty"`
+	NetTxBytes   uint64 `json:"net_tx_bytes,omitempty"`
+	NetTxErrors  uint64 `json:"net_tx_errors,omitempty"`
+	NetTxDropped uint64 `json:"net_tx_dropped,omitempty"`
 }
 
 // NewAgent ...
```
malnick: We should finish the comment before this merges.

collector/client.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	defer resp.Body.Close()
+
+	// Special case: on HTTP 401 Unauthorized, exit immediately. Considering this
+	//  could be running in a goroutine, we don't want the caller to fail forever.
+	if resp.StatusCode == http.StatusUnauthorized {
+		log.Fatalf("fetch error: unauthorized: please provide a suitable auth token using the AUTH_CREDENTIAL environment variable.")
```
rji: Nit: this is provided via `dcos-go/jwt` and/or `authToken`, not an environment variable.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
-	a.Topic = topic
-	a.AgentStateChan = make(chan *AgentState)
+	a.MetricsChan = metricsChan
+
+	// Detect the agent's IP address once. Per the DC/OS docs (at least as of
+	// November 2016), changing a node's IP address is not supported.
+	//
+	if a.AgentIP == "" { // allows us to mock this during testing
+		if a.AgentIP, err = a.getIP(); err != nil {
+			log.Error(err)
```
malnick: return err?

collector/client_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		c := NewHTTPClient(host, "/", time.Duration(1*time.Second))
+		err = c.Fetch(&data)
+
+		So(data["foo"], ShouldEqual, "bar")
+		So(err, ShouldBeNil)
+	})
+
+	// TODO(roger): write a test for the auth portion once it's implemented
+	Convey("Should fetch data AUTHENTICATED from the Mesos API and return", t, nil)
```
rji: TODO; unimplemented test.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+	var containers []agentContainer
+
+	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
+	c := NewHTTPClient(
+		strings.Join([]string{a.AgentIP, strconv.Itoa(a.Port)}, ":"),
+		"/containers",
+		time.Duration(15*time.Second))
+	if err := c.Fetch(&containers); err != nil {
+		log.Error(err)
+		return nil, err
```
malnick: I'm a big fan of returning the err right away and having the the top most caller log the err.Error()

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+// getAgentMetrics queries an agent for system-level metrics, such as CPU,
+// memory, disk, and network.
+//
+// TODO(roger): the metrics available via the /metrics/snapshot endpoint DO NOT
+// satisfy the requirements set forth in the original dcos-metrics design doc.
+// I've included the base Mesos agent metrics here as a starting point ONLY.
+// In the future, this should be replaced by something like
+// github.com/shirou/gopsutil.
+func (a *Agent) getAgentMetrics() (agentMetricsSnapshot, error) {
+	var metrics agentMetricsSnapshot
```
malnick: you could init the struct with {} here and then when returning below you can return metrics on err instead of re-initializing the struct later

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+
+	// TODO(roger): 5sec timeout is a guess. Is there a better way to do this?
+	c := NewHTTPClient(
+		strings.Join([]string{a.AgentIP, strconv.Itoa(a.Port)}, ":"),
+		"/metrics/snapshot",
+		time.Duration(5*time.Second))
+	if err := c.Fetch(&metrics); err != nil {
+		log.Error(err)
+		return agentMetricsSnapshot{}, err
```
malnick: return metrics, err ?

producers/http/http.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Infof("HTTP Producer serving requests on :%d", p.config.Port)
+		return http.ListenAndServe(fmt.Sprintf(":%d", p.config.Port), r)
+	}
+}
+
+// janitor analyzes the objects in the store and removes stale objects. An
+// object is considered stale when the top-level timestamp of its MetricsMessage
+// has exceeded the CacheExpiry, which is calculated as a multiple of the
+// collector's polling period. This function should be run in its own goroutine.
+func (p *producerImpl) janitor() {
```
rji: The janitor is a current source of races. Need to consider synchronizing calls to `p.store` by placing accesses into a single goroutine, and have multiple goroutines submit read/write requests for the store to a channel.

rji: Question: does that functionality belong in dcos-metrics, or dcos-go/store? :thinking:

producers/http/logger.go
```diff
@@ -14,4 +12,27 @@
...
+	log "github.com/Sirupsen/logrus"
+)
+
+func logger(inner http.Handler, name string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+
+		inner.ServeHTTP(w, r)
+
+		log.Printf(
```
rji: TODO: let's define a log level here instead of `Printf()`

producers/interface.go
```diff
@@ -0,0 +1,73 @@
...
+	Timestamp  int64       `json:"_timestamp"`
+}
+
+// Datapoint represents a single metric's timestamp, value, and unit in a response.
+// A single datapoint is typically contained in an array, such as []Datapoint{}.
+type Datapoint struct {
+	Name      string `json:"name"`
+	Value     string `json:"value"`
+	Unit      string `json:"unit"`
+	Timestamp string `json:"timestamp"` // time.RFC3339Nano, e.g. "2016-01-01T01:01:01.10000000Z"
```
rji: Nit: comment should be `time.RFC3339`, not `time.RFC3339Nano`

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
-	} else {
-		log.Printf("Failed to retrieve container metrics from agent at %s: %s", a.AgentIP, err)
+	return metrics, nil
+}
+
+// getAgentState fetches the state JSON from the Mesos agent, which contains
+// info such as framework names and IDs, the current leader, config flags,
+// container (executor) labels, and more.
+func (a *Agent) getAgentState() (agentState, error) {
+	var state agentState
```
malnick: same comment on init'ing the struct right away instead of on return.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+
+	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
+	c := NewHTTPClient(
+		strings.Join([]string{a.AgentIP, strconv.Itoa(a.Port)}, ":"),
+		"/state",
+		time.Duration(15*time.Second))
+	if err := c.Fetch(&state); err != nil {
+		log.Error(err)
+		return agentState{}, err
```
malnick: return state, err?

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
 
-// fetches container-level resource metrics from the agent (via /containers), emits to the framework topics (default 'metrics-<framework_id>')
-func (a *Agent) getContainerMetrics(agentState *AgentState, stats chan<- statsd.StatsEvent) ([]*AvroDatum, error) {
-	rootJSON, err := a.getJSONFromAgent("/containers", agentTestContainersFileFlag, stats)
-	if err != nil {
-		return nil, err
-	}
+// pollAgent queries the DC/OS agent for metrics and returns.
+func (a *Agent) pollAgent() metricsMeta {
+	var metrics metricsMeta
```
malnick: That pesky init struct comment again.

collector/agentPoller.go
```diff
@@ -15,451 +15,433 @@
...
+		if framework.ID == frameworkID {
+			return framework, true
 		}
 	}
-
-	return &AgentState{
-		agentID:          agentID,
-		frameworkNames:   frameworkNames,
-		executorAppNames: executorAppNames}, nil
+	return frameworkInfo{}, false
```
malnick: Init struct to var early and often! This can come back to bite you will nil values later. I always try to `var thing myStruct{}` if there is no data in it. Then you can return `thing` instead of init'ing a new struct later.

rji: I did address the other cases where I didn't initialize the struct early though (thanks for the feedback!)

collector/agentPoller.go
rji: Resolved in 5d14555.

producers/http/http_test.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			resp, err := http.Get(fmt.Sprintf("http://127.0.0.1:%d", port))
+			if err != nil {
+				panic(err)
+			}
+			defer resp.Body.Close()
+
+			b, err := ioutil.ReadAll(resp.Body)
+			fmt.Println(b)
```
rji: Need to actually test something here :)

collector/client.go
rji: Since this PR is specifically around the work needed to get the HTTP producer working, I'm inclined to leave this for a separate PR (to prevent this PR from getting any larger than it already is).

producers/http/http.go
rji: @malnick, @darkonie -- what are your thoughts on adding a goroutine to dcos-go/store that handles all reads/writes to the map, versus implementing it here? It's probably the kind of thing that we'd need elsewhere at some point.

rji: Talked to @malnick, going to go with a goroutine here.

PR #16, opened 2016-11-09
Add ping endpoint for reverse proxies

rji:
Can be used by reverse proxies to determine backend service health.


malnick:
:ship: it!

PR #17, opened 2016-11-09
TCP Listener

malnick:
Framework TCP listener.


malnick:
@rji @darkonie This successfully transforms the example framework metrics from Avro to our common `MetricsMessage{}` type. I'm looking for a good faith review of how I'm doing this transformation, since it feels a bit unorthodox. 

*Notes:*
- The TCP listener code and other methods for implementing a `io.Reader` still need to be refactored
- There are no unit tests and I've broken the tests on our main package
- It adds a switch statement that doesn't actually pass the transformed data to the HTTP producer yet

collector/framework_collector.go
```diff
@@ -0,0 +1,302 @@
...
+				Name:  fmt.Sprintf("%s", field.Fields[0].Datum),
+				Value: fmt.Sprintf("%s", field.Fields[1].Datum),
+				Unit:  fmt.Sprintf("%s", field.Fields[2].Datum),
+			}
+			datapoints = append(datapoints, dp)
+		}
+		pmm.Datapoints = datapoints
+	}
+
+	if fieldType == "" {
```
darkonie: maybe we can use `else if` here?
it looks like we are checking `fieldType` for empty value or `dcos.metrics.Tag` and `dcos.metrics.Datapoint`
what if we received something else?

collector/framework_collector.go
```diff
@@ -0,0 +1,302 @@
...
+			continue
+		}
+		recordsChan <- &AvroDatum{datum, topic, approxBytesRead}
+	}
+}
+
+func GetTopic(obj interface{}) (string, bool) {
+	record, ok := obj.(*goavro.Record)
+	if !ok {
+		return "UNKNOWN_RECORD_TYPE", false
```
darkonie: can we return error here instead of a string and handle the error in the caller?

PR #18, opened 2016-11-12
Full implementation for host-level metrics

rji:
This PR includes the full implementation for host-level metrics, via https://github.com/shirou/gopsutil. It also contains several fixes based on feedback during an internal demo, including:

* Renames agentPoller.go to agent.go. Eventually, there will also be a master.go, and the higher level polling logic should be moved into its own, new file, poller.go.
* Stopped including the full metric prefix in the datapoint name and updated tests to catch this.
* Leading dot problem should now be resolved. Added placeholders for functional tests for the `/containers` and `/containers/ {id}` endpoints, to be implemented at a later time.
* All datapoints' values now return numbers instead of strings.
* Removed "name" and "_timestamp" fields from the JSON output; they were only meant for internal use.
* Modified `/containers` endpoint to only return IDs, since each container has its own endpoint.
* Removed single metrics handlers (cpu, mem, etc) since all of that info was already available in a single container's `datapoints` array.


rji:
JSON output from my dev environment:

[/system/metrics/api/v0/agent](https://github.com/dcos/dcos-metrics/files/587521/agent.json.txt)
[/system/metrics/api/v0/containers](https://github.com/dcos/dcos-metrics/files/587520/containers.json.txt)
[/system/metrics/api/v0/containers/70997c43-4970-4bb2-aa31-9307829eb705](https://github.com/dcos/dcos-metrics/files/587522/containers_70997c43-4970-4bb2-aa31-9307829eb705.json.txt)

malnick:
:+1:

PR #24, opened 2016-11-16
Use Real Metric Values

malnick:
- [x] Type assert our way to real values for the framework metrics
- [x] Pass MesosID, ClusterID, Hostname & IPAddress into the node collector dimensions

### DIFF
--- a/collector/avro_record.go
+++ b/collector/avro_record.go
@@ -36,6 +36,9 @@ func (ar avroRecord) extract(pmm *producers.MetricsMessage) error {
 	// Extract tags
 	if fieldType == "dcos.metrics.Tag" {
 		for _, field := range ar {
+			if len(field.Fields) != 2 {
+				return errors.New(fmt.Sprintf("Tags must have 2 fields, got %d", len(field.Fields)))
+			}
 			fwColLog.Debugf("Adding tag %s", field)
 			tagName := fmt.Sprintf("%v", field.Fields[0].Datum)
 			tagValue := fmt.Sprintf("%v", field.Fields[1].Datum)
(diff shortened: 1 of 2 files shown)


malnick:
@darkonie @rji This is ready for review.

collector/avro_record.go
```diff
@@ -58,12 +58,49 @@ func (ar avroRecord) extract(pmm *producers.MetricsMessage) error {
...
 			fwColLog.Debugf("Adding datapoint %s", field)
+
 			dp := producers.Datapoint{
 				Name:      fmt.Sprintf("%v", field.Fields[0].Datum),
-				Value:     fmt.Sprintf("%v", field.Fields[1].Datum),
 				Unit:      fmt.Sprintf("%v", field.Fields[2].Datum),
 				Timestamp: fmt.Sprintf("%v", time.Now()),
 			}
+
+			value := field.Fields[1].Datum
```
darkonie: can we check `field.Fields` length? this might be a runtime error if you don't have any elements in your slice

collector/avro_record.go
```diff
@@ -58,12 +58,51 @@ func (ar avroRecord) extract(pmm *producers.MetricsMessage) error {
...
+			switch value.(type) {
```
darkonie: this looks a little bit strange, why do you switch type if in any case you just assign `value` to `dp.Value`? All this switch is not needed. You can just `dp.Value = value` it will be the same. I think you are missing something here?

malnick: The value of `value.(type)` is the type I'm indexing for. In this way, we catch the correct value for `.(type)` in the statement which follows.

collector/avro_record.go
malnick: @darkonie This is now fixed.

darkonie:
🚢 much better

PR #25, opened 2016-11-17
Version Flag

malnick:
- Add version flag
- Add ldflags for overriding version & revision at build time


malnick:
@darkonie CC

darkonie:
🚢

rji:
Can we also update the version string in http_client.go to use this?

malnick:
I'm not seeing that string in that file.

PR #26, opened 2016-11-17
Reduce timeout for polling state

malnick:


producers/http/handlers.go
```diff
@@ -27,14 +27,14 @@ import (
...
 		if nodeMetrics, err := p.store.GetByRegex(producers.NodeMetricPrefix + ".*"); err != nil {
 			w.WriteHeader(http.StatusBadRequest)
 		} else {
-			for _, v := range nodeMetrics {
-				am = append(am, v)
-			}
-			encode(am[0], w)
+			//for _, v := range nodeMetrics {
+			//	am = append(am, v)
+			//}
```
malnick: @rji Why was this implemented with the array? This induced a few nil pointers when nodeMetrics is empty. It looks like you can pass nodeMetrics directly to writer with out any modifications and the result looks good.

malnick:
CC @darkonie

collector/mesos_agent.go
```diff
@@ -137,7 +137,7 @@ func (h *DCOSHost) getAgentState() (agentState, error) {
 	c := NewHTTPClient(
 		strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
 		"/state",
-		time.Duration(15*time.Second))
+		time.Duration(2*time.Second))
```
darkonie: is 2 seconds long enough? maybe increase it?

collector/node_collector.go
```diff
@@ -110,43 +112,39 @@ func (h *DCOSHost) RunPoller() {
...
+	if h.DCOSRole == dcos.RoleAgent {
 		// always fetch/emit agent state first: downstream will use it for tagging metrics
 		nodeColLog.Debugf("Fetching state from DC/OS host %s:%d", h.IPAddress, h.Port)
 		agentState, err := h.getAgentState()
+
 		if err != nil {
 			nodeColLog.Errorf("Failed to get agent state from %s:%d. Error: %s", h.IPAddress, h.Port, err)
-			return metrics
+		} else {
+			metrics.agentState = agentState
```
darkonie: if this ok if we don't update `metrics.agentState`? this looks odd, maybe we should return error or something?

collector/mesos_agent.go
malnick: It's enough, and blocks gathering metrics. Remember, we're getting this from the localhost, so anything more than 2 seconds and something is wrong.

collector/node_collector.go
```diff
@@ -110,43 +112,39 @@ func (h *DCOSHost) RunPoller() {
...
+	metrics.timestamp = now.Unix()
+
+	// Fetch node-level metrics for all DC/OS roles
+	nodeColLog.Debugf("Fetching node-level metrics from DC/OS host %s:%d", h.IPAddress, h.Port)
+	nm, err := h.getNodeMetrics()
+	if err != nil {
+		nodeColLog.Errorf("Failed to get node-level metrics. %s", err)
+	} else {
+		metrics.nodeMetrics = nm
+	}
```
malnick: Yes, we're gathering metrics from HTTP endpoints and host level stuff here. If one fails, we shouldn't block the others and we shouldn't bail out or panic. It's best effort.

collector/node_collector.go
malnick: Same as above.

malnick:
@darkonie I updated all the handlers, let me know if anything else needs further attention.

producers/http/handlers.go
```diff
@@ -24,34 +24,47 @@ import (
...
+		}
+
+		if len(am) != 0 {
 			encode(am[0], w)
+			return
 		}
+
+		httpLog.Error("/api/v0/node - no content in store.")
+		http.Error(w, "No values found in store", http.StatusBadRequest)
+		return
```
darkonie: 👍 looks good, no need to final return

producers/http/handlers.go
```diff
@@ -24,34 +24,47 @@ import (
...
-			encode(cm, w)
+		containerMetrics, err := p.store.GetByRegex(producers.ContainerMetricPrefix + ".*")
+		if err != nil {
+			httpLog.Error("/api/v0/containers - %s", err.Error())
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		for _, c := range containerMetrics {
+			cm = append(cm, c.(producers.MetricsMessage).Dimensions.ContainerID)
```
darkonie: i'd be tempted to type assert on a new line
```go
a, ok := c.(producers.MetricsMessage)
if !ok {
    return ...
}
```

this will ensure we don't panic if unable to type assert

darkonie:
@malnick nits, but overall looks much better 👍

producers/http/handlers.go
malnick: Fixed.

malnick:
Fixed type assertion handling, :ship:?

PR #31, opened 2016-11-18
HTTP Helpers / Cleaned Up Main Package

malnick:
- [ ] break out HTTP helpers from entry point file
- [ ] add TLS configuration 
- [ ] break out config from entry point file


config.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return nil
+}
+
+// newConfig establishes our default, base configuration.
+func newConfig() Config {
+	return Config{
+		Collector: CollectorConfig{
+			HTTPProfiler:  true,
+			IPCommand:     "/opt/mesosphere/bin/detect_ip",
```
darkonie: i dont think you need this now

config.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+	c.IPAddress = ip.String()
+
+	id, err := nodeInfo.MesosID(nil)
+	if err != nil {
+		c.MesosID = ""
+	} else {
+		c.MesosID = id
+	}
```
darkonie: if you don't handle error maybe this would be better to read
```go
c.MesosID, err = nodeInfo.MesosID(nil)
if err != nil {
  logrus.Errorf("Could not get mesos ID: %s", err)
}
```

darkonie:
👍

malnick:
Thank you sir, comments addressed. Will merge after tests pass.

PR #35, opened 2016-11-19
Use a generalized http.Client

malnick:
- [ ] Refactor http_client to only have a single Fetch() method which accepts a http.Client
- [ ] Set client in main config using the generalized `http_helpers.NewMetricsClient()`
- [ ] Put http_helpers.go in their own package so we can leverage it in other sub packages

### DIFF
--- a/collector/http_client_test.go
+++ b/collector/http_client_test.go
@@ -20,17 +20,10 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"testing"
-	"time"
 
+	httpHelpers "github.com/dcos/dcos-metrics/http_helpers"
 	. "github.com/smartystreets/goconvey/convey"
 )
-
-func TestNewHTTPClient(t *testing.T) {
-	Convey("Should return a new client", t, func() {
-		c := NewHTTPClient("foo.example.com", "/bar", time.Duration(1*time.Second))
-		So(c, ShouldHaveSameTypeAs, &HTTPClient{})
-	})
-}
 
 func TestHTTPClient_Fetch(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -47,6 +40,11 @@ func TestHTTPClient_Fetch(t *testing.T) {
 	}))
 	defer ts.Close()
 
+	testClient, err := httpHelpers.NewMetricsClient("", "")
+	if err != nil {
+		t.Error("Error retreiving HTTP Client:", err)
+	}
+
 	Convey("Should fetch data UNAUTHENTICATED from the Mesos API and return", t, func() {
 		var data map[string]string
 		host, err := extractHostFromURL(ts.URL)
@@ -54,8 +52,12 @@ func TestHTTPClient_Fetch(t *testing.T) {
 			panic(err)
 		}
 
-		c := NewHTTPClient(host, "/", time.Duration(1*time.Second))
-		err = c.Fetch(&data)
+		testURL := url.URL{
+			Scheme: "http",
+			Host:   host,
+			Path:   "/",
+		}
+		err = Fetch(testClient, testURL, &data)
 
 		So(data["foo"], ShouldEqual, "bar")
 		So(err, ShouldBeNil)
@@ -63,13 +65,6 @@ func TestHTTPClient_Fetch(t *testing.T) {
 
 	// TODO(roger): write a test for the auth portion once it's implemented
 	Convey("Should fetch data AUTHENTICATED from the Mesos API and return", t, nil)
-}
-
-func TestHTTPClient_URL(t *testing.T) {
-	Convey("Should return the URL as a string", t, func() {
-		c := NewHTTPClient("foo.example.com", "/bar", time.Duration(1*time.Second))
-		So(c.URL(), ShouldEqual, "http://foo.example.com/bar")
-	})
 }
 
 func extractHostFromURL(u string) (string, error) {
--- a/collector/mesos_agent.go
+++ b/collector/mesos_agent.go
@@ -15,6 +15,7 @@
 package collector
 
 import (
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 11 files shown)


malnick:
@darkonie @rji - Please have a look when you get a moment.

@darkonie - Can you take a look at my sharing of Transport in particular? I am pretty sure I only create 2 transports, one for getting NodeInfo{} and another is passed in the http.Client to the collector. I don't want to run into the same leak you had in re-using connections :) 

@rji - This gets the JWT transport stuff into the collector package using a single, generalized client and shared transport instance (see comment above). We still need set configuration for the protocol used (http vs https) in the collector package.

collector/mesos_agent.go
```diff
@@ -114,17 +115,15 @@ type resourceStatistics struct {
...
-	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
-	c := NewHTTPClient(
-		strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
-		"/containers",
-		time.Duration(2*time.Second))
-	if err := c.Fetch(&containers); err != nil {
-		return nil, err
+	u := url.URL{
+		Scheme: "http",
+		Host:   strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
```
darkonie: use `net.JoinHostPort`

collector/mesos_agent.go
```diff
@@ -114,17 +115,15 @@ type resourceStatistics struct {
...
-
-	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
-	c := NewHTTPClient(
-		strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
-		"/containers",
-		time.Duration(2*time.Second))
-	if err := c.Fetch(&containers); err != nil {
-		return nil, err
+	u := url.URL{
+		Scheme: "http",
```
darkonie: if you run in EE shouldn't it be `https`?

collector/mesos_agent.go
```diff
@@ -114,17 +115,15 @@ type resourceStatistics struct {
...
+		Host:   strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
+		Path:   "/containers",
 	}
 
-	return containers, nil
+	h.HTTPClient.Timeout = time.Duration(2 * time.Second)
```
darkonie: but still it is better to extract the constant

collector/mesos_agent.go
```diff
@@ -133,16 +132,15 @@ func (h *DCOSHost) getContainerMetrics() ([]agentContainer, error) {
...
-	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
-	c := NewHTTPClient(
-		strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
-		"/state",
-		time.Duration(2*time.Second))
-	if err := c.Fetch(&state); err != nil {
-		return state, err
+	u := url.URL{
+		Scheme: "http",
+		Host:   strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
```
darkonie: `net.JoinHostPort`

collector/mesos_agent.go
```diff
@@ -133,16 +132,15 @@ func (h *DCOSHost) getContainerMetrics() ([]agentContainer, error) {
...
 
-	// TODO(roger): 15sec timeout is a guess. Is there a better way to do this?
-	c := NewHTTPClient(
-		strings.Join([]string{h.IPAddress, strconv.Itoa(h.Port)}, ":"),
-		"/state",
-		time.Duration(2*time.Second))
-	if err := c.Fetch(&state); err != nil {
-		return state, err
+	u := url.URL{
+		Scheme: "http",
```
darkonie: http vs https in EE?

collector/http_client_test.go
```diff
@@ -47,15 +40,24 @@ func TestHTTPClient_Fetch(t *testing.T) {
 	}))
 	defer ts.Close()
 
+	testClient, err := httpHelpers.NewMetricsClient("", "")
+	if err != nil {
+		t.Error("Error retreiving HTTP Client:", err)
```
darkonie: toy should probably avoid `t.Error` in tests. Use `t.Fatal` or `t.Fatalf`

collector/mesos_agent_test.go
```diff
@@ -92,6 +93,11 @@ func TestGetContainerMetrics(t *testing.T) {
 	}))
 	defer ts.Close()
 
+	testClient, err := httpHelpers.NewMetricsClient("", "")
+	if err != nil {
+		t.Error("Error retreiving HTTP Client:", err)
```
darkonie: t.Fatalf("Error retreiving HTTP Client: %s", err)

collector/node_collector_test.go
```diff
@@ -108,6 +117,11 @@ func TestNewDCOSHost(t *testing.T) {
 }
 
 func TestTransform(t *testing.T) {
+	testClient, err := httpHelpers.NewMetricsClient("", "")
+	if err != nil {
+		t.Error("Error retreiving HTTP Client:", err)
```
darkonie: t.Fatal

collector/mesos_agent.go
malnick: See PR comment, that's next. Requires larger refactor in configuration.

collector/mesos_agent.go
malnick: Same as above.

collector/http_client_test.go
malnick: why?

collector/mesos_agent_test.go
malnick: why?

malnick:
Thanks @darkonie - only question is regarding t.Fatal vs. Error.

rji:
`t.Fatal()` logs the failure and fails the test execution immediately, whereas `t.Error()` logs the failure but continues execution. Up to this point we've been using `t.Fatal()` in order to "fail fast".

PR #37, opened 2016-11-20
Decompose Collector Config 

malnick:
- Decompose collectors into individual packages
- Move configuration to globally set in main package
- Add `request_protocol` setting to node collector config
- Refactor unit tests to follow new collector packing pattern


malnick:
@darkonie This PR will fix the proto schemes eventually.

malnick:
Our integration tests are broken until Monday due to some reverted changes (see github.com/dcos/dcos/pull/778) but I can confirm this change works as expected in manual testing.

rji:
I hadn't split these up into individual packages yet because there's still a fair amount of code reuse among the collectors. For example, I don't think we need two different pollers here (one for the "agent" and one for the "node" metrics, and in the future, potentially a third for getting master metrics? all of these could be a single poller goroutine), nor do I think we need two implementations of `buildDatapoints`.

rji:
Does `request_protocol` actually need to be a configurable option? Mesos supports HTTPS already; this seems like the kind of thing where we could try HTTPS first and if we can't connect then fallback to HTTP and log a warning.

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +149,109 @@ type resourceStatistics struct {
...
+	for {
+		select {
+		case _ = <-ticker.C:
+			h.pollMesosAgent()
+			for _, m := range h.transformContainerMetrics() {
+				h.MetricsChan <- m
+			}
+		}
+	}
+}
```
darkonie:
```go
for {
  h.pollMesosAgent()
  for _, m := range h.transformContainerMetrics() {
      h.MetricsChan <- m
  }
  time.Sleep(time.Duration(h.PollPeriod) * time.Second)
}
```

replace time.Sleep with select if you need to use Ticker

darkonie: if you need to use channels, because in your next refactor you want to interrupt the waiting time by an event, you can do
```go
  select {
    case <-ticker.C:
  }
```

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +149,109 @@ type resourceStatistics struct {
...
+
+func (h *MesosAgentCollector) pollMesosAgent() {
+	now := time.Now().UTC()
+	h.timestamp = now.Unix()
+
+	host := fmt.Sprintf("%s:%d", h.NodeInfo.IPAddress, h.Port)
```
darkonie: use `net.JoinHostPort`

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +149,109 @@ type resourceStatistics struct {
...
+
+	// always fetch/emit agent state first: downstream will use it for tagging metrics
+	maColLog.Debugf("Fetching state from DC/OS host %s", host)
+	if err := h.getAgentState(); err != nil {
+		maColLog.Errorf("Failed to get agent state from %s. Error: %s", host, err)
```
darkonie: this is not a very good pattern when you do not return on error. I'd rather split this into smaller functions and return on error. This would be much better for unittesting and code readability

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +149,109 @@ type resourceStatistics struct {
...
+	return http_client.Fetch(h.HTTPClient, u, &h.containerMetrics)
 }
 
 // getAgentState fetches the state JSON from the Mesos agent, which contains
 // info such as framework names and IDs, the current leader, config flags,
 // container (executor) labels, and more.
-func (h *DCOSHost) getAgentState() (agentState, error) {
-	state := agentState{}
+func (h *MesosAgentCollector) getAgentState() error {
+	h.agentState = agentState{}
```
darkonie: maybe `&agentState{}`? so you won't have to return `&h.agentState`

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +149,109 @@ type resourceStatistics struct {
...
+			MesosID:            h.NodeInfo.MesosID,
+			ClusterID:          h.NodeInfo.ClusterID,
+			Hostname:           h.NodeInfo.Hostname,
+			ContainerID:        c.ContainerID,
+			ExecutorID:         c.ExecutorID,
+			FrameworkID:        c.FrameworkID,
+			FrameworkName:      fi.Name,
+			FrameworkRole:      fi.Role,
+			FrameworkPrincipal: fi.Principal,
+			Labels:             getLabelsByContainerID(c.ContainerID, h.agentState.Frameworks),
```
darkonie: `getLabelsByContainerID` should probably return error and you need to handle the error before you initiate `msg.Dimensions`

collector/node/node_collector_test.go
```diff
@@ -0,0 +1,88 @@
...
+		panic(err)
+	}
+
+	Convey("When building a slice of producers.Datapoint for a MetricsMessage", t, func() {
+		Convey("Should return the node's datapoints with valid tags and values", func() {
+			result := buildDatapoints(mockNodeMetrics, testTime)
+			So(len(result), ShouldEqual, 46)
+			So(result[0].Name, ShouldEqual, "uptime")
+			So(result[0].Unit, ShouldEqual, "")            // TODO(roger): no easy way to get units
+			So(result[0].Value, ShouldEqual, uint64(7865)) // TODO(roger): everything is a string for MVP
```
rji: This TODO no longer applies

rji:
I didn't really see an issue with keeping all of this as a single `collector` package for now (at least when I wrote it), but if we're going to go down this route then perhaps it's best to define an interface for each of the collectors to implement. All of this seems a bit premature to me though, especially for the initial release / MVP. Just my $0.02.

dcos-metrics.go
```diff
@@ -63,42 +63,33 @@ func main() {
 	if producerIsConfigured("http", cfg) {
 		log.Info("HTTP producer enabled")
 		c := httpProducer.Config{
-			Port:        cfg.Producers.HTTPProducerConfig.Port,
-			DCOSRole:    cfg.DCOSRole,
-			CacheExpiry: time.Duration(cfg.Collector.PollingPeriod) * time.Second * 2,
+			Port:     cfg.Producers.HTTPProducerConfig.Port,
+			DCOSRole: cfg.DCOSRole,
+			// TODO(malnick) constant or from config
+			CacheExpiry: time.Minute * 10,
```
malnick: Not really, I forget why I made this change. I can revert it.

malnick:
Yeah, I agree that a full blown interface for this isn't required just yet. The reason I went down the break out road now is so we can configure the individual collectors more easily, and not copy around configuration from main (setting a bunch of config to a new instance of a collector from our top level config that happens to be using subsets of collector config) and instead unmarshling directly to the configuration for the collector and then not touching it again. That optimization was a bit premature but simplifies breaking this out quite a bit.

collector/mesos_agent/mesos_agent.go
malnick: True, and I considered breaking them into separate funcs. However, we would only print the error and pass, then run the next individual method if we broke them out. Since the behavior is identical in both case, I left it as-is.

collector/mesos_agent/mesos_agent.go
malnick: It returns the return from Fetch() which requires you to pass reference of the struct which it unmarshal's JSON to.

collector/node/node_collector.go
```diff
@@ -0,0 +1,169 @@
...
+	MetricsChan chan producers.MetricsMessage
+	NodeInfo    collector.NodeInfo
+
+	nodeMetrics nodeMetrics
+	timestamp   int64
+}
+
+// RunPoller periodiclly polls the HTTP APIs of a Mesos agent. This function
+// should be run in its own goroutine.
+func (h *NodeCollector) RunPoller() {
```
malnick: Yeah, they both implement the same entrypoint func.

malnick:
@rji Fixed the cache expiry in last commit, removed an unneccessary copy of config to new config in main for http producer and removed the yaml tag for IP in the http producer config.

collector/mesos_agent/mesos_agent.go
malnick: I want to use select and ticker over sleep here.

collector/mesos_agent/mesos_agent.go
```diff
@@ -115,36 +148,109 @@ type resourceStatistics struct {
...
 }
 
+func (h *MesosAgentCollector) RunPoller() {
+	ticker := time.NewTicker(h.PollPeriod * time.Second)
+
+	// Poll once immediately
+	h.pollMesosAgent()
+	for _, m := range h.transformContainerMetrics() {
+		h.MetricsChan <- m
+	}
```
malnick: Ah I see now

malnick: Updated

collector/mesos_agent/mesos_agent.go
```diff
@@ -174,3 +280,69 @@ func getLabelsByContainerID(containerID string, frameworks []frameworkInfo) map[
...
+		case reflect.Struct:
+			pts = append(pts, buildDatapoints(f.Interface(), t)...)
+			continue
+		}
+
+		// Get the underlying value; see https://golang.org/pkg/reflect/#Kind
+		var uv interface{}
+		switch f.Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			uv = f.Int()
```
malnick: I don't want to touch that code in this refactor, I'm not really sure how it works yet. Ok to leave it for this PR?

collector/node/node_collector.go
```diff
@@ -0,0 +1,169 @@
...
+	ticker := time.NewTicker(h.PollPeriod * time.Second)
+
+	// Poll once immediately
+	h.pollHost()
+	for _, m := range h.transform() {
+		h.MetricsChan <- m
+	}
```
darkonie: same here, you can remove lines 49-52

collector/mesos_agent/mesos_agent.go
rji: This is required to get the underlying value in its native type. See the first law of reflection: https://blog.golang.org/laws-of-reflection

rji:
I've given this a few passes already and CI is still passing, so I'm going to say :ship:

collector/mesos_agent/mesos_agent.go
rji: That would be the case, except all of the structs define types for the fields. The playground example you provided assumes that the field has type `interface{}`. The functional tests in the repo, the Laws of Reflection blog post, and this example (https://play.golang.org/p/O-LZlNqY5I) show how this works.

collector/node/node_collector.go
```diff
@@ -0,0 +1,169 @@
...
+		case reflect.Struct:
+			pts = append(pts, buildDatapoints(f.Interface(), t)...)
+			continue
+		}
+
+		// Get the underlying value; see https://golang.org/pkg/reflect/#Kind
+		var uv interface{}
+		switch f.Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			uv = f.Int()
```
rji: Not true, see previous comment.

PR #38, opened 2016-11-22
Make golint and govet work in CI

rji:
golint and govet were working for local development, but weren't working reliably in CI. Hopefully this fixes it.


rji:
This also resolves a number of issues found by golint and govet in order to get a green build.

malnick:
:+1:

collector/mesos_agent/mesos_agent.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package mesos_agent
```
malnick: Why are we renaming this package? It's common convention in golang to underscore, not camelcase, the package name: https://blog.golang.org/package-names

config.go
```diff
@@ -50,9 +50,9 @@ type Config struct {
 // portion of this project. That is, the code responsible for querying Mesos,
 // et. al to gather metrics and send them to a "producer".
 type CollectorConfig struct {
-	HTTPProfiler bool                             `yaml:"http_profiler"`
-	Node         *node.NodeCollector              `yaml:"node,omitempty"`
-	MesosAgent   *mesos_agent.MesosAgentCollector `yaml:"mesos_agent,omitempty"`
```
malnick: See underscore vs. package name comment above. For variables, camel case is fine. However, for files and package names the convention is lower case with underscores.

http_helpers/http_helpers.go
```diff
@@ -1,4 +1,4 @@
-package http_helpers
```
malnick: Naming convention again, this should be underscores and lowercase

schema/generator.go
```diff
@@ -132,7 +132,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't open output %s: %s", *outfileFlag, err)
 	}
-	if _, err = fmt.Fprint(outfile, `package metrics_schema
```
malnick: Same here.

malnick:
Please see https://blog.golang.org/package-names for package naming conventions. Another good resource is net/http, which has a similar sub package structure as this project. The naming should be lowercase and underscores instead of camel case.

rji:
@malnick -- the package names have been changed based on recommendations from [golang/golint](https://github.com/golang/lint). The blog post you linked to is over a year old. According to the latest documentation at https://golang.org/doc/effective_go.html#package-names:

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps. [...] Another convention is that the package name is the base name of its source directory; the package in src/encoding/base64 is imported as "encoding/base64" but has name base64, not encoding_base64 and not encodingBase64.

By that convention, both of us are incorrect. But I believe the [MixedCase section of the Effective Go docs](https://golang.org/doc/effective_go.html#mixed-caps) are the tie breaker here:

> Finally, the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.

Golint complains about underscores in package names, and not mixedCase. For consistency, this project has implemented all recommendations from golint up to this point; I'd like to continue that.

rji:
Specifically: https://github.com/golang/lint/blob/206c0f020eba0f7fbcfbc467a5eb808037df2ed6/lint.go#L538-L540 (golang/lint at 206c0f0, which is the SHA for HEAD as of this comment)

rji:
Alternately, I'm happy to push up a commit that downcases all package names if you don't like mixedCase. That should make the linter pass as well.

malnick:
The current post you have won me over, mixed caps it is.

rji:
Cool, let me change up the filenames to match the package names for consistency. Then this will be ready for review/merge.

rji:
@malnick -- I think you might like this one :) none of the packages are mixedCase nor do they contain underscores now :smile:

rji:
Ah, looks like it needs a rebase though.

malnick:
Love it, :ship: it.

PR #40, opened 2016-11-23
(fixup) s/swam/swap/

rji:
Oops!

@malnick -- Let me know if you'd like a test to validate these names as well, or if codifying them here is enough.


malnick:
:ship:

PR #45, opened 2016-11-30
Test versioning

ghost:
Adds tests around binary versioning and the provided User-Agent string for the built-in HTTP client.

### DIFF
--- a/util/http/client/client_test.go
+++ b/util/http/client/client_test.go
@@ -56,7 +56,10 @@ func TestClient(t *testing.T) {
 			panic(err)
 		}
 
-		So(gotUserAgent, ShouldContainSubstring, "com.mesosphere.dcos-metrics/")
+		// Default is "unset"; this is set by 'scripts/build.sh' via ldflags.
+		// The actual test for the version in the built binary is in TestMain,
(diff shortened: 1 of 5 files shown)


util/http/client/client_test.go
```diff
@@ -25,7 +27,40 @@ import (
...
+		u, err := url.Parse(ts.URL)
+		if err != nil {
+			panic(err)
+		}
+
+		if err := Fetch(http.DefaultClient, *u, nil); err != nil {
+			panic(err)
+		}
+
+		So(gotUserAgent, ShouldContainSubstring, "com.mesosphere.dcos-metrics/")
```
rji: @malnick -- I'd like to do something similar to what you did with the linker flags for setting the `userAgent` string, but that appears to only happen on a `go build` and not during `go test`. I can easily modify the ldflags in the build script, but I'm not sure how I'd actually test it. Do you have any ideas here?

rji: I suppose a hacky way of testing that the User-Agent was modified successfully during the build process would be to expose the user agent when calling `-version` on the command line, but I'd like to avoid that if I can.

PR #46, opened 2016-11-30
Validate license headers

ghost:
* Fails the build if a `*.go` source file doesn't contain the Apache license header and copyright statement. Ignores files in `./vendor/**` and `./examples/**`.
* Adds missing headers to source files.


rji:

PR #49, opened 2016-12-01
Degrade if cert path is not found

malnick:
If a CA cert is not present but path is given, we need to degrade to no cert verification. The method proposed in this commit will also make it so if cert verification fails, we degrade to no cert verification. I am not sure if this is the best way to go about this. We might want to consider a helper method that attempts loading the cert from a path in another method, and degrading only if the certificate path that is given does not exist, and not if the cert verification or other failures that might occur in the load CA cert method happen.

### DIFF
--- a/util/http/helpers/helpers.go
+++ b/util/http/helpers/helpers.go
@@ -55,7 +55,12 @@ func getTransport(caCertificatePath string) (*http.Transport, error) {
 		log.Infof("Loading CA cert: %s", caCertificatePath)
 		caPool, err := loadCAPool(caCertificatePath)
 		if err != nil {
-			return tr, err
+			log.Errorf("Error loading CA: %s", err.Error())
+			log.Warn("Skipping certificate verification.")
+			tr.TLSClientConfig = &tls.Config{
+				InsecureSkipVerify: true,
+			}
+			return tr, nil
 		}
 
 		tr.TLSClientConfig = &tls.Config{


malnick:
@rji @alberts - I found an issue where the CA cert is unavailable on public agents, the service fails to start since it bails out if the cert path given isn't found. The method proposed here is a hot fix, and I want to get some dialogue about how to handle this situation, or examples of how we handle it in other code.

rji:
> If a CA cert is not present but path is given, we need to degrade to no cert verification. The method proposed in this commit will also make it so if cert verification fails, we degrade to no cert verification.

I disagree. In my opinion, if the user specified a path to a CA cert, they've requested that the certs be verified. If they didn't provide a path to a CA chain, then we skip verification. Therefore, if the certs couldn't be loaded (e.g., the path doesn't exist), it's considered fatal, and this service should fail to start (which was the behavior prior to this PR).

Further, if the user has specified a CA chain and wants certificates to be verified, it should be an error if verification failed. This service can't simply detect that verification failed and continue anyway; otherwise, what's the point of doing the verification at all?

I'll note that I haven't proposed any solutions here, I'm not quite sure what the next step forward would be. But I don't think the behavior in this PR enhances security or follows the principle of least astonishment. Citing a previous implementation in this project: if we attempt to load the configuration file for this service via the `-config` command line flag, and the file doesn't exist, this service fails to start. I'd expect the same to be true here.

Overall, -1 to this patch, but happy to discuss alternate solutions.

rji:
Getting back to the root of the problem though, why isn't the CA chain present on public agents?

malnick:
I'm not sure, and I actually think it might be failed logic in my bootstrap code on EE. I'm working with @alberts to figure it out. I agree with your logic here, and don't want to degrade if a specified path isn't present.

PR #50, opened 2016-12-02
Default Majority of Configuration 

malnick:
Default the configuration we are not passing in via the config.yaml for the DC/OS installer. We should only override things in enterprise config file with EE specific config, everything else can be defaulted for now. TODO - expose collector polling times in config.yaml.

@rji When you get a moment can you review this?

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,12 +18,16 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"strings"
+	"time"
 
 	"github.com/dcos/dcos-go/dcos"
 	"github.com/dcos/dcos-go/dcos/nodeutil"
 	mesosAgent "github.com/dcos/dcos-metrics/collectors/mesos/agent"
 	"github.com/dcos/dcos-metrics/collectors/node"
 	httpProducer "github.com/dcos/dcos-metrics/producers/http"
+	httpClient "github.com/dcos/dcos-metrics/util/http/client"
 	httpHelpers "github.com/dcos/dcos-metrics/util/http/helpers"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -92,7 +92,8 @@ func (c *Config) setFlags(fs *flag.FlagSet) {
 func (c *Config) loadConfig() error {
 	fileByte, err := ioutil.ReadFile(c.ConfigPath)
 	if err != nil {
-		return err
+		log.Warnf("%s not found. Using all defaults", c.ConfigPath)
```
malnick: Can do this or place logic around `len(c.ConfigPath)` and if it's `> 0` then we try to load config. I'm open to both use cases.

rji:
@malnick -- I pushed a commit that makes additional changes needed to get this working properly. I also need to write some tests to ensure this doesn't change in the future. I'll take ownership on this PR and ping you when the tests are in place.

ghost (approved):
Hey @malnick -- I made a few changes to your original PR and added some tests. When you get a chance, could you give this a once-over?

config.go
```diff
@@ -174,14 +177,34 @@ func getNewConfig(args []string) (Config, error) {
...
+		fmt.Printf(strings.Join([]string{
+			"DC/OS Metrics Service",
+			fmt.Sprintf("Version: %s", VERSION),
+			fmt.Sprintf("Revision: %s", REVISION),
+			fmt.Sprintf("HTTP User-Agent: %s", httpClient.USERAGENT),
+		}, "\n"))
+		os.Exit(0)
+	}
+
+	if len(c.ConfigPath) > 0 {
```
malnick: :+1:

malnick:
I had one minor comment about adding the `role` to the version output. Other than that lets :ship: it.

rji:
Awesome, thank you sir! I'll add that to the output.

malnick:
Oh, one more thing, we're probably only accessing the version command from the CLI - why do we get the user agent from HTTP client?

rji:
Yeah, I don't like the user agent output with `-version`, but I do that so that I can test that the linker is replacing the User-Agent string in the build script, and that the final binary doesn't have an "unset" value. I wasn't sure of a nicer way to do this in testing, short of replacing the linker flags with a call out to sed :frowning_face:. If you have any suggestions on how to improve that though, I'd love to hear them!

malnick:
Haha, no this is fine.

PR #51, opened 2016-12-06
Added INJECT_TAG_X_Y environment variables

susanxhuynh:
Added INJECT_TAG_X_Y environment variables which allow users to inject arbitrary tags via the config.

Example: I have a Datadog instance which monitors multiple clusters. This patch allows me to add my own "cluster.name:my-dcos-cluster" tag to distinguish between different clusters, by setting INJECT_TAG_CLUSTER_NAME=my-dcos-cluster.

cc @nickbp for review


rji:
We already do this on the collector side. These consumers haven't been updated though. At some point they will be rewritten.

PR #57, opened 2016-12-14
RC1 bug fixes and code cleanup

rji:
This PR resolves issues found while testing RC1, specifically around node info being copied / set in individual collectors instead of once when the service starts up. Along with fixing that, this PR also includes a bunch of collector code cleanup to make each collector package more consistent.

There's also a new `collectors.MetricsCollector` interface, but it'll take a bit more work to get the collectors to use it. I'll save that for a future PR.


collectors/node/node.go
```diff
@@ -26,65 +26,72 @@ import (
...
 	nodeMetrics nodeMetrics
 	timestamp   int64
 }
 
+// New returns a new instance of the node metrics collector and a metrics chan.
+func New(cfg Collector, nodeInfo collectors.NodeInfo) (Collector, chan producers.MetricsMessage) {
+	c := Collector{
+		PollPeriod:  cfg.PollPeriod,
+		MetricsChan: make(chan producers.MetricsMessage),
+		log:         log.WithFields(log.Fields{"collector": "node"}),
```
malnick: :+1:  I really like adding logger to the `New` instance

collectors/framework/framework.go
```diff
@@ -60,100 +64,42 @@ type countingReader struct {
...
-	if err := datapointData.createObjectFromRecord(dps); err != nil {
-		return pmm, err
-	}
-
-	if err := datapointData.extract(&pmm); err != nil {
-		return pmm, err
-	}
-
-	return pmm, nil
+func New(cfg Collector, nodeInfo collectors.NodeInfo) (Collector, chan producers.MetricsMessage) {
```
malnick: Really happy to see this as a standalone method.

collectors/mesos/agent/agent.go
```diff
@@ -36,25 +36,20 @@ const (
...
 	RequestProtocol string        `yaml:"request_protocol"`
-
-	MetricsChan chan producers.MetricsMessage
-	NodeInfo    collectors.NodeInfo
-	HTTPClient  *http.Client
+	HTTPClient      *http.Client
 
 	agentState       agentState
 	containerMetrics []agentContainer
+	log              *logrus.Entry
```
malnick: Nit: For people who may be unfamiliar, if might be nice to drop in a comment to documentation `https://godoc.org/github.com/sirupsen/logrus#Entry`

rji: Can do!

PR #59, opened 2016-12-19
Fetch leader /state endpoint for framework principal

rji:
Prior to this commit, we only fetched state from the local agent, but
only the leader's /state endpoint can provide the framework principal.

This commit adds a method to query the master's /state endpoint for the
framework principal. It also adds a mocked cluster state object to
agent_test.go, and ensures that FrameworkPrincipal is set to the correct
value.

### DIFF
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -46,6 +46,7 @@ type Collector struct {
 	HTTPClient      *http.Client
 
 	agentState       agentState
+	clusterState     clusterState
 	containerMetrics []agentContainer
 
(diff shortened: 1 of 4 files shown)


collectors/mesos/agent/agent.go
```diff
@@ -46,6 +46,7 @@ type Collector struct {
 	HTTPClient      *http.Client
 
 	agentState       agentState
+	clusterState     clusterState
```
malnick: nit (not blocking): masterState vs clusterState variable name

malnick:
LGTM :ship:

PR #61, opened 2017-01-18
Node collector: unit values, nodeMetricPoller interface refactor 

malnick:
- [x] Fix unit names
- [x] Enable tagging metrics
- [x] Update unit tests
- [x] Push logic of metric transformation down to the methods which gather the metrics rather than the generic transform method in the top level node collection

## For reviewers
This is a very large PR. I've structured the commits in a logical order to make this easier to digest: 
- Adding unit values to the metrics 
- Adding `tags` parameter to the `producers.MetricMessage{}` interface
- Enabling the usage of a generic `nodeMetricPoller` type for metric gathering
- Then, adding a file and the associated unit test for each metric type

This commit history should be clean enough to work through commit-by-commit, and I suggest it be done that way for sanity sake. TIA. 

## Integration
PR to DC/OS: https://github.com/dcos/dcos/pull/1155

### DIFF
--- a/collectors/node/cpu.go
+++ b/collectors/node/cpu.go
@@ -1,0 +1,152 @@
+package node
+
+import (
+	"math"
+	"sync"
+
+	"github.com/dcos/dcos-metrics/producers"
+	"github.com/shirou/gopsutil/cpu"
+)
+
+type cpuCoresMetric struct {
+	cpuCores  int32
+	cpuTotal  float64
+	cpuUser   float64
+	cpuSystem float64
+	cpuIdle   float64
+	cpuWait   float64
+	timestamp string
+}
+
+func (m *cpuCoresMetric) poll() error {
+	ts := thisTime()
+	hi, err := getNumCores()
+	if err != nil {
+		return err
+	}
+
+	cpuStatePcts := calculatePcts(getCPUTimes())
+
+	m.cpuCores = hi
+	m.timestamp = ts
+	m.cpuTotal = cpuStatePcts.User + cpuStatePcts.System
+	m.cpuUser = cpuStatePcts.User
+	m.cpuSystem = cpuStatePcts.System
+	m.cpuIdle = cpuStatePcts.Idle
+	m.cpuWait = cpuStatePcts.Iowait
+
+	return nil
+}
+
+func (m *cpuCoresMetric) addDatapoints(nc *nodeCollector) error {
+	cpuDps := []producers.Datapoint{
+		producers.Datapoint{
+			Name:      CPU_CORES,
+			Unit:      COUNT,
+			Value:     m.cpuCores,
+			Timestamp: m.timestamp,
+		},
+		producers.Datapoint{
+			Name:      CPU_TOTAL,
+			Unit:      COUNT,
+			Value:     m.cpuTotal,
+			Timestamp: m.timestamp,
+		},
+		producers.Datapoint{
+			Name:      CPU_USER,
+			Unit:      COUNT,
+			Value:     m.cpuUser,
+			Timestamp: m.timestamp,
+		},
+		producers.Datapoint{
+			Name:      CPU_SYSTEM,
+			Unit:      COUNT,
+			Value:     m.cpuSystem,
+			Timestamp: m.timestamp,
+		},
+		producers.Datapoint{
+			Name:      CPU_IDLE,
+			Unit:      COUNT,
+			Value:     m.cpuIdle,
+			Timestamp: m.timestamp,
+		},
+		producers.Datapoint{
+			Name:      CPU_WAIT,
+			Unit:      COUNT,
+			Value:     m.cpuWait,
+			Timestamp: m.timestamp,
+		},
+	}
+
+	for _, dp := range cpuDps {
+		nc.datapoints = append(nc.datapoints, dp)
+	}
+
+	return nil
+}
+
+// -- helpers
+func getNumCores() (int32, error) {
+	cores := int32(0)
+	cpus, err := cpu.Info()
+	if err != nil {
+		return cores, err
+	}
+	for _, c := range cpus {
+		cores += c.Cores
+	}
+	return cores, nil
+}
+
+/* CPU Helper Methods */
+type lastCPUTimes struct {
+	sync.Mutex
+	times cpu.TimesStat
+}
+
+var lastCPU lastCPUTimes
+
+func init() {
+	t, _ := cpu.Times(false) // get totals, not per-cpu stats
+	lastCPU.Lock()
+	lastCPU.times = t[0]
+	lastCPU.Unlock()
+}
+
+func getCPUTimes() (cpu.TimesStat, cpu.TimesStat) {
+	currentTimes, _ := cpu.Times(false) // get totals, not per-cpu stats
+	lastTimes := lastCPU.times
+
+	lastCPU.Lock()
+	lastCPU.times = currentTimes[0] // update lastTimes to the currentTimes
+	lastCPU.Unlock()
+
+	return currentTimes[0], lastTimes
+}
+
+// calculatePct returns the percent utilization for CPU states. 100.00 => 100.00%
+func calculatePcts(lastTimes cpu.TimesStat, curTimes cpu.TimesStat) cpu.TimesStat {
+	totalDelta := curTimes.Total() - lastTimes.Total()
+	if totalDelta == 0 {
+		totalDelta = 1 // can't divide by zero
+	}
+	return cpu.TimesStat{
+		User:      round(math.Dim(curTimes.User, lastTimes.User) / totalDelta * 100),
+		System:    round(math.Dim(curTimes.System, lastTimes.System) / totalDelta * 100),
+		Idle:      round(math.Dim(curTimes.Idle, lastTimes.Idle) / totalDelta * 100),
+		Nice:      round(math.Dim(curTimes.Nice, lastTimes.Nice) / totalDelta * 100),
+		Iowait:    round(math.Dim(curTimes.Iowait, lastTimes.Iowait) / totalDelta * 100),
+		Irq:       round(math.Dim(curTimes.Irq, lastTimes.Irq) / totalDelta * 100),
+		Softirq:   round(math.Dim(curTimes.Softirq, lastTimes.Softirq) / totalDelta * 100),
+		Steal:     round(math.Dim(curTimes.Steal, lastTimes.Steal) / totalDelta * 100),
+		Guest:     round(math.Dim(curTimes.Guest, lastTimes.Guest) / totalDelta * 100),
+		GuestNice: round(math.Dim(curTimes.GuestNice, lastTimes.GuestNice) / totalDelta * 100),
+		Stolen:    round(math.Dim(curTimes.Stolen, lastTimes.Stolen) / totalDelta * 100),
+	}
+}
+
+// Helper function for rounding to two decimal places
+func round(f float64) float64 {
+	shift := math.Pow(10, float64(2))
+	return math.Floor(f*shift+.5) / shift
+}
--- a/collectors/node/cpu.go
+++ b/collectors/node/cpu.go
@@ -1,3 +1,17 @@
+// Copyright 2016 Mesosphere, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package node
 
 import (
--- a/collectors/node/cpu.go
+++ b/collectors/node/cpu.go
@@ -39,7 +39,12 @@ func (m *cpuCoresMetric) poll() error {
 		return err
 	}
 
-	cpuStatePcts := calculatePcts(getCPUTimes())
+	currentTime, lastTime, err := getCPUTimes()
+	if err != nil {
+		return err
+	}
+
+	cpuStatePcts := calculatePcts(currentTime, lastTime)
 
 	m.cpuCores = hi
 	m.timestamp = ts
@@ -92,10 +97,7 @@ func (m *cpuCoresMetric) addDatapoints(nc *nodeCollector) error {
 		},
 	}
 
-	for _, dp := range cpuDps {
-		nc.datapoints = append(nc.datapoints, dp)
-	}
-
+	nc.datapoints = append(nc.datapoints, cpuDps...)
 	return nil
 }
 
@@ -127,15 +129,19 @@ func init() {
 	lastCPU.Unlock()
 }
 
-func getCPUTimes() (cpu.TimesStat, cpu.TimesStat) {
-	currentTimes, _ := cpu.Times(false) // get totals, not per-cpu stats
+func getCPUTimes() (cpu.TimesStat, cpu.TimesStat, error) {
+	currentTimes, err := cpu.Times(false) // get totals, not per-cpu stats
+	if err != nil {
+		return cpu.TimesStat{}, cpu.TimesStat{}, err
+	}
+
 	lastTimes := lastCPU.times
 
 	lastCPU.Lock()
 	lastCPU.times = currentTimes[0] // update lastTimes to the currentTimes
 	lastCPU.Unlock()
 
-	return currentTimes[0], lastTimes
+	return currentTimes[0], lastTimes, nil
 }
 
 // calculatePct returns the percent utilization for CPU states. 100.00 => 100.00%
--- a/collectors/node/cpu.go
+++ b/collectors/node/cpu.go
@@ -57,8 +57,8 @@ func (m *cpuCoresMetric) poll() error {
 	return nil
 }
 
-func (m *cpuCoresMetric) addDatapoints(nc *nodeCollector) error {
-	cpuDps := []producers.Datapoint{
+func (m *cpuCoresMetric) getDatapoints() ([]producers.Datapoint, error) {
+	return []producers.Datapoint{
 		producers.Datapoint{
 			Name:      CPU_CORES,
 			Unit:      COUNT,
@@ -95,10 +95,7 @@ func (m *cpuCoresMetric) addDatapoints(nc *nodeCollector) error {
 			Value:     m.cpuWait,
 			Timestamp: m.timestamp,
 		},
-	}
-
-	nc.datapoints = append(nc.datapoints, cpuDps...)
-	return nil
+	}, nil
 }
 
 // -- helpers
--- a/collectors/node/filesystem.go
+++ b/collectors/node/filesystem.go
@@ -1,3 +1,17 @@
+// Copyright 2016 Mesosphere, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package node
 
 import (
--- a/collectors/node/filesystem.go
+++ b/collectors/node/filesystem.go
@@ -1,0 +1,115 @@
+package node
+
+import (
+	"github.com/dcos/dcos-metrics/producers"
+	"github.com/shirou/gopsutil/disk"
+)
+
+type filesystemMetrics struct {
+	fsMetrics []filesystemMetric
+}
+
+type filesystemMetric struct {
+	path        string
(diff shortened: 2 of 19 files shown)


collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
 			continue
 		}
 
+		// Get the unit for the metric
+		var un = ""
```
darkonie: nit: `var un string`

collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
...
+		uSlice := strings.Split(parsed.String(), ".")
+		uName := uSlice[len(uSlice)-1]
```
darkonie: this will panic if slice is empty

collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
...
+			}
+
+		default:
+			un = "count"
+		}
+
 		pts = append(pts, producers.Datapoint{
 			Name:      parsed.String(),
-			Unit:      "", // TODO(roger): not currently an easy way to get units
+			Unit:      un, // TODO(roger): not currently an easy way to get units
```
malnick: nit: legacy comment

collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
 			continue
 		}
 
+		// Get the unit for the metric
+		var un = ""
+		uSlice := strings.Split(parsed.String(), ".")
+		uName := uSlice[len(uSlice)-1]
+		mName := uSlice[0]
```
darkonie: this is also not safe, can we validate uSlice len before doing this?

collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
...
+			switch uName {
+			case "bytes":
+				un = "bytes"
+			default:
+				un = "count"
+			}
+
+		default:
+			un = "count"
+		}
```
darkonie: maybe instead of nested switch statements, we can make functions to extract the values base on input parameter? it'll be easier to unittest and read

collectors/node/node.go
```diff
@@ -162,9 +162,54 @@ func (c *Collector) buildDatapoints(in interface{}, t time.Time) []producers.Dat
...
+		var un = ""
+		uSlice := strings.Split(parsed.String(), ".")
+		uName := uSlice[len(uSlice)-1]
+		mName := uSlice[0]
+
+		switch mName {
+		case "memory":
+
+			switch uName {
+			case "free", "buffers", "cached":
```
darkonie: please extract the constants

collectors/node/node.go
darkonie: maybe instead of `un`, we can use a more meaningful variable?

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+			Name:      CPU_WAIT,
+			Unit:      COUNT,
+			Value:     m.cpuWait,
+			Timestamp: m.timestamp,
+		},
+	}
+
+	for _, dp := range cpuDps {
+		nc.datapoints = append(nc.datapoints, dp)
+	}
```
darkonie: nit: easier to read
```
nc.datapoints = append(nc.datapoints, cpuDps...)
```

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+// -- helpers
+func getNumCores() (int32, error) {
+	cores := int32(0)
+	cpus, err := cpu.Info()
+	if err != nil {
+		return cores, err
+	}
+	for _, c := range cpus {
+		cores += c.Cores
+	}
```
darkonie: nit: easier to read
```
return len(c.Cores), nil
```

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+	return cores, nil
+}
+
+/* CPU Helper Methods */
+type lastCPUTimes struct {
+	sync.Mutex
+	times cpu.TimesStat
+}
+
+var lastCPU lastCPUTimes
```
darkonie: can we try to avoid global variables?

collectors/node/metrics_test.go
```diff
@@ -193,22 +130,22 @@ func TestGetCPUTimes(t *testing.T) {
...
-func TestGetNetworkInterfaces(t *testing.T) {
-	Convey("When getting network interfaces", t, func() {
-		Convey("Should return a list containing nodeNetworkInterface{} structs", func() {
-			ifs := getNetworkInterfaces()
-			So(len(ifs), ShouldBeGreaterThan, 0)
-			So(ifs, ShouldHaveSameTypeAs, []nodeNetworkInterface{})
-		})
-	})
-}
+//func TestGetFilesystems(t *testing.T) {
```
malnick: Cleaned up this commented out code in https://github.com/dcos/dcos-metrics/pull/61/commits/57db8bc140eafe2cdddbcb2ff3bb2fa57c880b60

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+
+func init() {
```
darkonie: let's not use global variables and do this logic once in a dedicated function.

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+	t, _ := cpu.Times(false) // get totals, not per-cpu stats
+	lastCPU.Lock()
+	lastCPU.times = t[0]
+	lastCPU.Unlock()
+}
+
+func getCPUTimes() (cpu.TimesStat, cpu.TimesStat) {
+	currentTimes, _ := cpu.Times(false) // get totals, not per-cpu stats
```
darkonie: let's not ignore `err` and return it to a caller?

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+	lastTimes := lastCPU.times
+
+	lastCPU.Lock()
+	lastCPU.times = currentTimes[0] // update lastTimes to the currentTimes
```
darkonie: i trust the library, however i'd also check for `currentTimes` length before accessing it's index

collectors/node/cpu.go
malnick: it's used by two funcs, I tried to refactor that out. I'm using the legacy code in this instance.

collectors/node/filesystem.go
```diff
@@ -0,0 +1,129 @@
...
+	m.fsMetrics = f
+	return nil
+}
+
+func (m *filesystemMetrics) addDatapoints(nc *nodeCollector) error {
+	/* Enumerate each filesystem found and add a datapoint object contining the
+	capacity and inode metrics plus a tag denoting the filesystem
+	path from which these came */
+	for _, fs := range m.fsMetrics {
+		nc.datapoints = append(nc.datapoints, producers.Datapoint{
```
darkonie: passing a structure reference to a function and then mutating it is not the best approach in go. Much better would be to pass required parameters and then return the result, the caller should mutate the required fields. Otherwise you'll see a lot of nasty bugs.

collectors/node/filesystem.go
```diff
@@ -0,0 +1,129 @@
...
+			Name:      FS_INODE_FREE,
+			Unit:      COUNT,
+			Value:     fs.inodesFree,
+			Timestamp: fs.timestamp,
+			Tags: map[string]string{
+				"path": fs.path,
+			},
+		})
+	}
```
darkonie: since you are changing a lot of fields here, maybe it'll make more sense to split it info smaller functions?

collectors/node/load.go
```diff
@@ -0,0 +1,70 @@
...
+// limitations under the License.
+
+package node
+
+import (
+	"github.com/dcos/dcos-metrics/producers"
+	"github.com/shirou/gopsutil/load"
+)
+
+type loadMetric struct {
```
darkonie: why not use `AvgStat` from https://github.com/shirou/gopsutil/blob/master/load/load.go#L15 ?

collectors/node/load.go
```diff
@@ -0,0 +1,70 @@
...
+	ts := thisTime()
+	l, err := load.Avg()
+	if err != nil {
+		return err
+	}
+
+	m.load1Min = l.Load1
+	m.load5Min = l.Load5
+	m.load15Min = l.Load15
+	m.timestamp = ts
```
darkonie: you could embed `AvgStat` to your structure + the timestamp, so you could update the load average fields with one assignment
https://github.com/shirou/gopsutil/blob/master/load/load.go#L15

collectors/node/load.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+func (m *loadMetric) addDatapoints(nc *nodeCollector) error {
```
darkonie: same here, it's better to make this function to return the result rather then make it modify the structure reference.

collectors/node/memory.go
```diff
@@ -0,0 +1,122 @@
...
+
+type memoryMetric struct {
+	memTotal   uint64
+	memFree    uint64
+	memBuffers uint64
+	memCached  uint64
+	swapTotal  uint64
+	swapFree   uint64
+	swapUsed   uint64
+	timestamp  string
```
darkonie: you could reuse the memory structures from https://github.com/shirou/gopsutil/blob/master/mem/mem.go#L19

collectors/node/memory.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (m *memoryMetric) poll() error {
+	ts := thisTime()
+	m.timestamp = ts
```
darkonie: nit: i think the timestamp must be the very last step in the function to make it more accurate

collectors/node/memory.go
```diff
@@ -0,0 +1,122 @@
...
+			Name:      SWAP_USED,
+			Unit:      BYTES,
+			Value:     m.swapUsed,
+			Timestamp: m.timestamp,
+		},
+	}
+
+	for _, dp := range memDps {
+		nc.datapoints = append(nc.datapoints, dp)
+	}
```
darkonie: same here, do not modify the reference, return the result

collectors/node/memory.go
```diff
@@ -0,0 +1,122 @@
...
+
+func getMemory() (*mem.VirtualMemoryStat, error) {
+	m, err := mem.VirtualMemory()
+	return m, err
+}
+
+func getSwap() (*mem.SwapMemoryStat, error) {
+	s, err := mem.SwapMemory()
+	return s, err
+}
```
darkonie: these functions are redundant 

`getSwap()` == `mem.SwapMemory()`

collectors/node/metrics.go
```diff
@@ -15,225 +15,105 @@
...
 }
 
-var lastCPU lastCPUTimes
-
-func init() {
-	t, _ := cpu.Times(false) // get totals, not per-cpu stats
-	lastCPU.Lock()
-	lastCPU.times = t[0]
-	lastCPU.Unlock()
+type nodeMetricPoller interface {
```
darkonie: any reason you don't export the interface?

collectors/node/metrics.go
```diff
@@ -15,225 +15,105 @@
...
-	}
-}
+// -- helpers
 
-// Helper function for rounding to two decimal places
-func round(f float64) float64 {
-	shift := math.Pow(10, float64(2))
-	return math.Floor(f*shift+.5) / shift
+func thisTime() string {
+	return time.Now().UTC().Format(time.RFC3339Nano)
```
darkonie: maybe it is a good idea to pass an argument to this function which will tell what format you want to use? and default to `time.RFC3339Nano` if none

collectors/node/network.go
```diff
@@ -0,0 +1,153 @@
...
+	netIn         uint64
+	netOut        uint64
+	netInPackets  uint64
+	netOutPackets uint64
+	netInDropped  uint64
+	netOutDropped uint64
+	netInErrors   uint64
+	netOutErrors  uint64
+	timestamp     string
+}
```
darkonie: you could reuse the structures from https://github.com/shirou/gopsutil/blob/master/net/net.go

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+
+type cpuCoresMetric struct {
+	cpuCores  int32
+	cpuTotal  float64
+	cpuUser   float64
+	cpuSystem float64
+	cpuIdle   float64
+	cpuWait   float64
+	timestamp string
+}
```
darkonie: reuse the structures from https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (m *cpuCoresMetric) poll() error {
```
darkonie: in every `poll()` function you update the structure with new data, but they are not thread (goroutine) safe. You should probably lock the data for read/write before you doing it.

collectors/node/metrics.go
malnick: It's only used internally to transform the metrics into our `producers.Datapoint` type.

collectors/node/cpu.go
```diff
@@ -0,0 +1,166 @@
...
+	m.cpuTotal = cpuStatePcts.User + cpuStatePcts.System
+	m.cpuUser = cpuStatePcts.User
+	m.cpuSystem = cpuStatePcts.System
+	m.cpuIdle = cpuStatePcts.Idle
+	m.cpuWait = cpuStatePcts.Iowait
+
+	return nil
+}
+
+func (m *cpuCoresMetric) addDatapoints(nc *nodeCollector) error {
```
darkonie: do not modify the reference to a structure, return the result and let the caller handle it.
Also, this approach isn't thread safe.

collectors/node/metrics.go
darkonie: let's figure out how to improve your interface. I think if we could make a robust interface, the implementation would be also different. Can you explain why do you need to use the interface here in first place?

collectors/node/network.go
```diff
@@ -0,0 +1,153 @@
...
+			Value:     nic.netOutErrors,
+			Timestamp: nic.timestamp,
+			Tags: map[string]string{
+				"interface": nic.interfaceName,
+			},
+		})
+	}
+
+	return nil
+}
```
darkonie: i have the same comments here, reuse the structure, use lock.Mutax, do not update reference to a structure.

collectors/node/uptime.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (m *uptimeMetric) addDatapoints(nc *nodeCollector) error {
+	nc.datapoints = append(nc.datapoints, producers.Datapoint{
+		Name:      UPTIME,
+		Unit:      COUNT,
+		Value:     m.uptime,
+		Timestamp: m.timestamp,
+	})
+	return nil
+}
```
darkonie: i have the same comments here, reuse the structure, use lock.Mutax, do not update reference to a structure.

collectors/node/cpu.go
malnick: We have to `range` over cpus to get `c`.

collectors/node/load.go
malnick: I want to save this for a later refactor in lieu of keeping this one from getting even larger :)

collectors/node/metrics.go
malnick: I am using it in an effort to set a precedence on how to add new metrics to this codebase later. If myself or another developer adds code to this, I want there to be a strictly defined pattern on how that is done. Though an interface may seem overly complex for this, since we can easily just add getMeomry() or getThisMetric() functions in lieu of them, the interface does provide a roadmap on how to add code to this project. 

If we use utility functions, we'll end up with different naming mechanisms, polling methods,  etc. Using an interface means every metric gathered is done so in the same way across the codebase. This makes it efficient to onboard new devs or to remind old devs how to add to this codebase if they haven't in a while. 

The extra 4 lines of code it took us to define this interface really define a roadmap, making this code less complex, even if on the surface it might not seem that way.

collectors/node/metrics.go
malnick: +1 if we need that later, for now all timestamps are the same.

collectors/node/network.go
malnick: See comment above regarding the complexity of this refactor.

malnick:
Fixed reference structure update issues and addressed most other comments.

collectors/node/cpu.go
```diff
@@ -127,15 +129,19 @@ func init() {
 	lastCPU.Unlock()
 }
 
-func getCPUTimes() (cpu.TimesStat, cpu.TimesStat) {
-	currentTimes, _ := cpu.Times(false) // get totals, not per-cpu stats
+func getCPUTimes() (cpu.TimesStat, cpu.TimesStat, error) {
+	currentTimes, err := cpu.Times(false) // get totals, not per-cpu stats
+	if err != nil {
+		return cpu.TimesStat{}, cpu.TimesStat{}, err
```
darkonie: nit: consider it in future refactor
we should probably return a reference to `cpu.TimeStat{}` instead of actual structure. That way you can return `nil`s on error. Also, you are making hard copies of `cpu.TimeStat{}` here.

collectors/node/filesystem.go
malnick: +1 for a another PR?

collectors/node/load.go
malnick: +1 for future PR?

collectors/node/cpu.go
```diff
@@ -0,0 +1,169 @@
...
+	m.cpuTotal = cpuStatePcts.User + cpuStatePcts.System
+	m.cpuUser = cpuStatePcts.User
+	m.cpuSystem = cpuStatePcts.System
+	m.cpuIdle = cpuStatePcts.Idle
+	m.cpuWait = cpuStatePcts.Iowait
+
+	return nil
+}
+
+func (m *cpuCoresMetric) getDatapoints() ([]producers.Datapoint, error) {
```
darkonie: can this function return an error?

collectors/node/cpu.go
darkonie: ok i see,
what about this function? https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go#L62

darkonie: can we use it instead?

collectors/node/cpu.go
malnick: No, but sometimes it can, and is required by the interface.

malnick: And in the future it might.

collectors/node/metrics.go
```diff
@@ -15,225 +15,107 @@
...
-		})
+	// For each metric poller defined, execute .poll() to get the latest
+	// metric, check for errors, and iterate over the slice of producers.Datapoint
+	// returned by .poll(), adding them to our top scope nodeMetrics slice.
+	for _, mp := range nodeMetricPollers {
+		if err := mp.poll(); err != nil {
+			return nc.datapoints, err
+		}
+
+		if dps, err := mp.getDatapoints(); err != nil {
```
darkonie: it looks like you can join `poll` and `getDatapoints` into one function, i don't see much value in having 2 separate functions. Considering the amount of work this might take to change the interface and tight timeframe, this can be addressed in the follow up PR

collectors/node/metrics.go
```diff
@@ -15,225 +15,107 @@
...
-			InodesTotal:        usage.InodesTotal,
-			InodesUsed:         usage.InodesUsed,
-			InodesFree:         usage.InodesFree,
-		})
+	// For each metric poller defined, execute .poll() to get the latest
+	// metric, check for errors, and iterate over the slice of producers.Datapoint
+	// returned by .poll(), adding them to our top scope nodeMetrics slice.
+	for _, mp := range nodeMetricPollers {
+		if err := mp.poll(); err != nil {
+			return nc.datapoints, err
```
darkonie: usually you don't want to return partial result on error, nil is a valid list you can return it
`return nil, err`

collectors/node/metrics.go
```diff
@@ -15,225 +15,107 @@
...
+		}
+
+		if dps, err := mp.getDatapoints(); err != nil {
+			return nc.datapoints, err
```
darkonie: `return nil, err`

collectors/node/metrics.go
```diff
@@ -15,225 +15,107 @@
...
+		} else {
```
darkonie: else is redundant here

malnick: dps is in the call space of the if statement

collectors/node/metrics.go
malnick: I need to check the err or return, else add dps to the datapoint slice.

collectors/node/metrics.go
malnick: They're two discrete actions, in one case I'm polling for methods and creating data. In another I'm transforming them to a slice of producers.Datapoint. this allows better unit testing and follows the form and function of "do one thing and do it well".

collectors/node/metrics.go
malnick: I do want a partial result here, the metrics service should error on the backend with a message but it needs to be a best effort to show as many datapoints requested as possible. 500'ing b/c a metric gathering poll failed shouldn't take out the entire service.

darkonie: 👍

PR #67, opened 2017-02-23
Fix for CPU percentages bug 

philipnrmn:
Usage stats were previously coerced to zero or more with `math.Dim` before
being the percentage calcuation was made. As they are based on a syscall
which yields processor ticks, the delta between old and new data was
always negative, thus they were always 0.

I tested with various loads locally and finally by replacing the `dcos-metrics` binary on a DC/OS cluster. 

Here's a snippet from the `/node` endpoint with the system running a single CPU-intensive process:
```
[
  {
    name: "cpu.total",
    value: 25.299999999999997,
    unit: "percent",
    timestamp: "2017-02-23T23:55:07.867637559Z"
  },
  {
    name: "cpu.user",
    value: 5.74,
    unit: "percent",
    timestamp: "2017-02-23T23:55:07.867637559Z"
  },
  {
    name: "cpu.system",
    value: 19.56,
    unit: "percent",
    timestamp: "2017-02-23T23:55:07.867637559Z"
  },
  {
    name: "cpu.idle",
    value: 74.38,
    unit: "percent",
    timestamp: "2017-02-23T23:55:07.867637559Z"
  },
  {
    name: "cpu.wait",
    value: 0.09,
    unit: "percent",
    timestamp: "2017-02-23T23:55:07.867637559Z"
  }
]
```

Here with four instances of the same process:
```
[
  {
    name: "cpu.total",
    value: 99.07,
    unit: "percent",
    timestamp: "2017-02-24T00:01:07.905291688Z"
  },
  {
    name: "cpu.user",
    value: 20.41,
    unit: "percent",
    timestamp: "2017-02-24T00:01:07.905291688Z"
  },
  {
    name: "cpu.system",
    value: 78.66,
    unit: "percent",
    timestamp: "2017-02-24T00:01:07.905291688Z"
  },
  {
    name: "cpu.idle",
    value: 0,
    unit: "percent",
    timestamp: "2017-02-24T00:01:07.905291688Z"
  },
  {
    name: "cpu.wait",
    value: 0,
    unit: "percent",
    timestamp: "2017-02-24T00:01:07.905291688Z"
  }
]
```


philipnrmn:
@malnick sure, that's now done

malnick:
Upstream https://github.com/dcos/dcos/pull/1290

collectors/node/cpu.go
```diff
@@ -167,3 +167,11 @@ func round(f float64) float64 {
 	shift := math.Pow(10, float64(2))
 	return math.Floor(f*shift+.5) / shift
 }
+
+// Helper function: replace all negative numbers with zero
+func gtZero(n float64) float64 {
+	if math.Signbit(n) {
```
darkonie: is there a particular reason you need to use `math.Signbit` here?

collectors/node/cpu.go
```diff
@@ -148,17 +148,17 @@ func calculatePcts(lastTimes cpu.TimesStat, curTimes cpu.TimesStat) cpu.TimesSta
...
+		System:    gtZero(round((curTimes.System - lastTimes.System) / totalDelta * 100)),
+		Idle:      gtZero(round((curTimes.Idle - lastTimes.Idle) / totalDelta * 100)),
+		Nice:      gtZero(round((curTimes.Nice - lastTimes.Nice) / totalDelta * 100)),
+		Iowait:    gtZero(round((curTimes.Iowait - lastTimes.Iowait) / totalDelta * 100)),
+		Irq:       gtZero(round((curTimes.Irq - lastTimes.Irq) / totalDelta * 100)),
+		Softirq:   gtZero(round((curTimes.Softirq - lastTimes.Softirq) / totalDelta * 100)),
+		Steal:     gtZero(round((curTimes.Steal - lastTimes.Steal) / totalDelta * 100)),
+		Guest:     gtZero(round((curTimes.Guest - lastTimes.Guest) / totalDelta * 100)),
+		GuestNice: gtZero(round((curTimes.GuestNice - lastTimes.GuestNice) / totalDelta * 100)),
+		Stolen:    gtZero(round((curTimes.Stolen - lastTimes.Stolen) / totalDelta * 100)),
```
darkonie: this looks sane to me, however, will that work properly on multi CPU / core hosts? Did you check?

collectors/node/cpu.go
philipnrmn: I thought it was better style than a simple less-than check. Would you prefer `n < 0`?

collectors/node/cpu.go
philipnrmn: Yes, I tested by running various loads locally and on a DC/OS cluster. I'll add some detail to the PR description.

collectors/node/cpu.go
darkonie: yes, Signbit shifts left 63 bits and check that the most significant bit is set. https://golang.org/src/math/signbit.go?s=233:261#L1

i don't think you really want this

darkonie: maybe it makes sense to move `n < 0` to helper function `round`? i don't think you ever need to round negative numbers?

philipnrmn: @darkonie addressed.

collectors/node/cpu.go
```diff
@@ -148,22 +148,25 @@ func calculatePcts(lastTimes cpu.TimesStat, curTimes cpu.TimesStat) cpu.TimesSta
...
+		GuestNice: roundAboveZero((curTimes.GuestNice - lastTimes.GuestNice) / totalDelta * 100),
+		Stolen:    roundAboveZero((curTimes.Stolen - lastTimes.Stolen) / totalDelta * 100),
 	}
 }
 
-// Helper function for rounding to two decimal places
-func round(f float64) float64 {
+// Helper function for rounding to two decimal places. Coerces negative numbers to zero.
+func roundAboveZero(f float64) float64 {
+	if f < 0 {
```
darkonie: `f < 1` ?

philipnrmn: No, this is definitely meant to be a zero check (it replaces the call to `math.Signbit`)

darkonie: so the expression
`return math.Floor(0*shift+.5) / shift` is valid? in this case 👍

collectors/node/cpu.go
```diff
@@ -148,22 +148,25 @@ func calculatePcts(lastTimes cpu.TimesStat, curTimes cpu.TimesStat) cpu.TimesSta
...
+		Guest:     roundAboveZero((curTimes.Guest - lastTimes.Guest) / totalDelta * 100),
+		GuestNice: roundAboveZero((curTimes.GuestNice - lastTimes.GuestNice) / totalDelta * 100),
+		Stolen:    roundAboveZero((curTimes.Stolen - lastTimes.Stolen) / totalDelta * 100),
 	}
 }
 
-// Helper function for rounding to two decimal places
-func round(f float64) float64 {
+// Helper function for rounding to two decimal places. Coerces negative numbers to zero.
+func roundAboveZero(f float64) float64 {
```
darkonie: sorry to be a pain
but name `roundAboveZero` suggests f > 0, in our case it can be 0, right?

philipnrmn: Yeah, this method is not very readable to me, I feel like it's doing too much. I'll break it down a little.

PR #69, opened 2017-02-24
Datadog HTTP Plugin

malnick:
## Overview
Creates a basic plugin for datadog that uses the metrics service HTTP API.

## Plugin Documentation & Tests
- [x] How to make the systemd unit?
- [x] How to build locally?
- [ ] Unit tests for the datadog plugin

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -11,8 +11,9 @@ To develop a new plugin in preparation for a pull requet to this project:
 1. Make a go file named after your package `touch plugins/cool/cool.go`
 
 #### Get a new plugin.Plugin{}
-1. Create falgs: `myFalgs := []cli.Flag{}`
-1. `myPlugin := plugin.New(yourFlags)` -> `plugin.Plugin{}`
+1. Create flags: `myFalgs := []cli.Flag{}`
+1. `myPlugin := plugin.New(myFlags)` -> `plugin.Plugin{}`
+1. Use `myPlugin.App.Action` and pass it a [cli.ActionFunc()](https://github.com/urfave/cli/blob/master/app.go#L66) which has all the `main()` logic your plugin needs.
 1. Call `myPlugin.Metrics()` which returns a slice of `producers.MetricsMessage{}` from the metrics HTTP API. 
 
 At this point you can write what ever helper methods you need to transform these and send to your metrics aggregation or cloud hosted service. 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,0 +1,49 @@
+## DC/OS Metrics Service Plugins
+**NOTE** These plugins are considered experimental and are not supported by the DC/OS support team. If you're having issues, please feel free to file an issue to this repo (not github.com/dcos/dcos) or file a *Pull Request* (we love pull requests!).
+
+Though your mileage may vary, if there is a plugin missing that you need, please feel free to file an issue so we can poll what services our users want the most. 
+
+## Developing
+To develop a new plugin in preparation for a pull requet to this project:
+
+### Plugin Development
+1. Create a new package for your `cool` plugin `mkdir plugins/cool/`
+1. Make a go file named after your package `touch plugins/cool/cool.go`
+
+#### Get a new plugin.Plugin{}
+1. Create falgs: `myFalgs := []cli.Flag{}`
+1. `myPlugin := plugin.New(yourFlags)` -> `plugin.Plugin{}`
+1. Call `myPlugin.Metrics()` which returns a slice of `producers.MetricsMessage{}` from the metrics HTTP API. 
+
+At this point you can write what ever helper methods you need to transform these and send to your metrics aggregation or cloud hosted service. 
+
+### Unit Test Coverage
+1. Add a unit test (aim for 80% coverage)
+
+### Add Plugin to Build Pipeline
+#### Add `cool` plugin to the build pipeline
+```
+vi Makefile
+# add a line to build your plugin from scripts/build.sh - follow the already shown 
+pattern for this.
+
+vi scripts.build.sh
+# add in a function() for building your plugin. Follow the patterns in this script for 
+adding your function.
+```
+
+#### Add a README.md in your `cool` package explaining:
+```
+vi plugins/cool/README.md
+# The Cool Plugin
+This cool plugin allows you to plugin to cool stuff.
+
+# How To Run Me
+1. ...
+
+# Caveats
+I knew there were caveats!
+``` 
+
+#### Submit a Pull Request! 
+Ping @malnick on DC/OS community slack :)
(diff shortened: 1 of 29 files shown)


scripts/build.sh
```diff
@@ -48,6 +48,11 @@ function build_statsd-emitter {
     examples/statsd-emitter/main.go
 }
 
+function build_datadog_plugin {
+		go build -a -o ${BUILD_DIR}/dcos-metrics-${COMPONENT}-${GIT_REF} \
```
malnick: white space, yo.

plugins/README.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+1. Add a README.md in your `cool` package explaining:
+```
+vi plugins/cool/README.md
+# The Cool Plugin
+This cool plugin allows you to plugin to cool stuff.
+
+# How To Run Me
+1. ...
```
darkonie: this looks unfinished?

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+		cli.IntFlag{
+			Name:  "datadog-port",
+			Value: 8125,
+			Usage: "Datadog port",
+		},
+	}
+
+	ddp, err := plugin.New(ddPluginFlags)
+	handleErr(err)
```
darkonie: it seems that you call this function only once, maybe we should  get rid of it and make
`log.Fatalf("Unable to initialize a new plugin: %s", err)`

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+
+	ddp.Action = func(c *cli.Context) error {
+		datadogAgent := fmt.Sprintf("%s:%d", c.String("datadog-host"), c.Int("datadog-port"))
```
darkonie: use [net.JoinHostPort](https://golang.org/pkg/net/#JoinHostPort)?

spahl:
As we discussed IRL, I would like to not have provider (datadog, signalfx...) specific code inside of DC/OS itself. We have been discussing running this as tasks.

spahl:
or provisioned OOB if need be

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+		metricsPaths, err := plugin.MetricsEndpoints(c.String("dcos-role"))
+		if err != nil {
+			return err
+		}
+
+		for {
+			for _, path := range metricsPaths {
+				metricsURL := url.URL{
+					Scheme: c.String("metrics-proto"),
+					Host:   fmt.Sprintf("%s:%d", c.String("metrics-host"), c.Int("metrics-port")),
```
darkonie: use net.JoinHostPort

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+	c.Namespace = "dcos-metrics"
+
+	tags, err := buildTags(mm.Dimensions)
+	if err != nil {
+		return err
+	}
+
+	for _, dp := range mm.Datapoints {
+		name := strings.Join([]string{mm.Name, dp.Name}, producers.MetricNamespaceSep)
+		val, err := strconv.ParseFloat(fmt.Sprintf("%v", reflect.ValueOf(dp.Value).Interface()), 64)
```
darkonie: this looks odd...
you take the dp.Value, print it out as a string and then uses strconv to convert string to float64

what if you type assert? would it work? Something like
```go
valueFloat, ok := dp.Value.(float64)
if !ok {
  return fmt.Errorf("Unable to type assert value %v to float64", dp.Value)
}
```

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+	v := reflect.ValueOf(msg)
+	if v.Kind() != reflect.Struct {
+		return tags, fmt.Errorf("error: expected type struct, got %s", v.Kind().String())
+	}
+
+	for i := 0; i < v.NumField(); i++ {
+		// The producers.Dimensions struct contains a nested
+		// map[string]string for user-defined Labels
+		if v.Field(i).Kind() == reflect.Map {
+			for k, v := range v.Field(i).Interface().(map[string]string) {
```
darkonie: while you check `v.Field(i).Kind() == reflect.Map` you type assert to a map[string]string. If this happens to be a `map[string]interface{}` for some reason, this will panic in runtime.

darkonie: if you certain 100% there will be only map[string]string, ignore my comment

plugins/datadog/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+		tag := strings.Join([]string{fieldTag, fieldVal}, ".")
+		tags = append(tags, tag)
+	}
+
+	return tags, nil
+}
+
+func getDCOSMetrics(request *http.Request) (producers.MetricsMessage, error) {
+	log.Infof("Making request to %+v", request.URL)
+	client := http.Client{}
```
darkonie: `client := &http.Client{}`

plugins/plugin.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, f := range extFlags {
+		app.Flags = append(app.Flags, f)
+	}
+
+	return app, nil
+}
+
+/* Helpers for metrics HTTP API */
+func MetricsEndpoints(role string) ([]string, error) {
+	if role == "master" {
```
darkonie: let's import constants from https://github.com/dcos/dcos-go/blob/master/dcos/constants.go

plugins/plugin.go
```diff
@@ -0,0 +1,73 @@
...
+		return []string{
+			"/system/v1/metrics/v0/node",
+		}, nil
+	}
+
+	if role == "agent" {
```
darkonie: let's import constants from https://github.com/dcos/dcos-go/blob/master/dcos/constants.go

plugins/datadog/datadog.go
malnick: We have to coerce it to float64, that's the only type it accepts.

malnick:
@spahl If we build out of band are you ok that this code lives in this repo?

plugins/datadog/datadog.go
malnick: Full disclaimer, i ripped this off from other code as a hold over. This entire func() will go away eventually.

scripts/build.sh
malnick: Not done with docs yet.

plugins/datadog/datadog.go
malnick: But it's not guaranteed to be a float64, it could be an int or any other number. We force coerce it to fit the interface for the library.

plugins/README.md
malnick: That's just an example section of an example plugin.

malnick:
@darkonie All set!

PR #75, opened 2017-03-01
use /container/<container_id>/ for container metrics

malnick:


plugins/plugin.go
```diff
@@ -156,13 +157,51 @@ func (p *Plugin) setEndpoints() error {
...
+		if err != nil {
+			logrus.Errorf("Encountered error reading response body, %s", err.Error())
+			return err
+		}
+
+		if err := json.Unmarshal(body, &containers); err != nil {
+			return err
+		}
+
+		for _, c := range containers {
```
nickbp: Looks like this is only fetched once at startup? What if containers are added/removed?

malnick: If you call `.Metrics()` in your plugin this will call `getEndpoints()` again.

malnick: As long as you call `.Metrics()` over and over this should refresh.

malnick: https://github.com/dcos/dcos-metrics/blob/a3ebfb6ecca8f71e93625bc5cd7838b8e3ee5eb0/plugins/plugin.go#L111

nickbp: oic makes sense

PR #77, opened 2017-03-02
Fix network metric name constants

malnick:
https://jira.mesosphere.com/browse/DCOS-14186


malnick:
Ha, great catch. Let me fix those.

PR #78, opened 2017-03-07
Functional Plugin Library

malnick:
Adds functional options and simplifies API usage.

### DIFF
--- a/plugins/option.go
+++ b/plugins/option.go
@@ -1,0 +1,14 @@
+package plugin
+
+import "github.com/urfave/cli"
+
+type Option func(*Plugin) error
+
+func ExtraFlags(extraFlags []cli.Flag) Option {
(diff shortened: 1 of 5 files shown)


plugins/option.go
```diff
@@ -12,3 +38,61 @@ func ExtraFlags(extraFlags []cli.Flag) Option {
...
+func PollingInterval(i int) Option {
+	return func(p *Plugin) error {
+		p.PollingInterval = i
+		return nil
+	}
+}
+
+func MetricsProtocol(proto string) Option {
+	return func(p *Plugin) error {
+		if proto != "http" || proto != "https" {
```
malnick: Yup, it's bad logic :)

malnick:
@philipnrmn Fixed the logic around that option, can you do another pass and merge if you're happy?

PR #79, opened 2017-03-07
CONTRIBUTING.md

malnick:


plugins/CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+		cli.StringFlag{
+			Name:  "datadog-host",
+			Value: datadogHost,
+			Usage: "Datadog output hostname",
+		},
+...
+``` 
+
+#### 2. Define a connection function 
+This is for transformating metrics from the DC/OS metrics service `[]producers.MetricsMessage{}` type to your 3rd party provider. Your connector func must accept a `[]producers.MetricsMessage{}' type and a urfave/cli `*cli.Context{}` and return an `error`. The connector func is ran in a for loop on the interval specified by `plugin.Plugin.PollingInterval`, an example looks like:
```
philipnrmn: s/is ran/runs/

philipnrmn:
I'm force-merging this, the tests are passing but the status isn't being correctly reported.

PR #85, opened 2017-03-21
Created Dockerfile that can be used to run the build.

collinvandyck:
It installs the `tree` dependency. At some point in the future the build may not require `tree` at which point we can remove the Dockerfile and just instruct folks to use `golang:1.7`.

cc @philipnrmn

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,13 @@ like contributors to follow.
 If you have any questions on the below guidelines, please feel free to contact us by using
 any of the community resources listed in this project's [README.md][dcos-metrics-readme] file.
 
+## Building
+
+A Makefile is included with the project.
+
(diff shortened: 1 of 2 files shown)


philipnrmn:
Thanks for looking after this!

Can we help people get up and running by specifying the exact Docker command, eg

```
docker build . -t dcos-metrics
docker run dcos-metrics make
```

philipnrmn:
Actually, why do we need to install `tree`?

I don't think it's a hard dependency, see:
https://github.com/dcos/dcos-metrics/blob/master/scripts/build.sh#L67

collinvandyck:
We think that the `tree` dependency installs a dependency that makes the cpu/node unit test work (does not work with the `golang:1.7` image).  I'm going to close this and possibly create another PR with the proper dependency.

PR #87, opened 2017-03-23
Librato Plugin

collinvandyck:
This changeset introduces a new [Librato](https://metrics.librato.com) plugin.  It adapts to the current plugin framework and converts dcos-metrics metrics into Librato metrics and then sends them to Librato.

### DIFF
--- a/plugins/librato/measurement.go
+++ b/plugins/librato/measurement.go
@@ -1,0 +1,104 @@
+// Copyright 2017 Mesosphere, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"regexp"
+	"strconv"
+)
+
+// measurement represents each point in a time series
+type measurement struct {
+	Name  string            `json:"name"`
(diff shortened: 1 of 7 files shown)


collinvandyck:
NB: I have found an edge case where some logically equivalent tag names (clusterid vs cluster_id) are getting reported for some metrics because of duplication between incoming tags and the MetricsMessage dimensions overlapping. I will be looking into this shortly.

collinvandyck:
This changeset is in a good state to merge now.

philipnrmn:
Can we add a simple Readme for this plugin?

plugins/librato/measurement.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("Could not set value: %v", err)
+	}
+	m.Value = val
+	return nil
+}
+
+func (m *measurement) addTag(name string, value string) error {
+	matched, err := regexp.MatchString(`^[-.:_\w]+\z{1,64}$`, name)
+	if err != nil {
+		return fmt.Errorf("Error occurred matching tag name: %v", err)
```
darkonie: you can use `%s` for err, since it implements Stringer interface. %v is more expensive

collinvandyck:
Updated the changeset with a README (cc @philipnrmn) and fixed for divide by zero and string formats (cc @darkonie).

plugins/librato/measurement.go
```diff
@@ -38,6 +38,10 @@ func newMeasurement() *measurement {
 }
 
 func (m *measurement) floorTime(interval int64) {
+	if interval <= 0 {
+		log.Errorf("Attempted to floor measurement time with interval %d", interval)
+		return
```
darkonie: not sure if it's a good idea to just return,
maybe if interval <= 0, you can assume it's = 1? otherwise you probably should return error to the caller

collinvandyck: I think i was erring (pun not intended.... or is it???) on the side of getting measures to the remote service if at all possible.  In this case, it's not a huge deal to not floor the time, as its only purpose is really to align timestamps in librato as much possible.  So a log.Errorf seemed reasonable tiven that.

PR #88, opened 2017-03-23
Syncs the YAML to docsite YAML, which was adjusted for optimal rendering

emanic:
@malnick @darkonie @philipnrmn This PR syncs the dcos-metrics YAML with the docsite YAML. I had to adjust that YAML for better rendering by adding tags. In addition, the docsite renderer displayed a red error with the previous YAML. The introductory content in this YAML is included elsewhere on the site so not needed within the YAML itself. Let me know if this looks OK to you, it will make things easier if we can keep the sources in sync. Thanks!

### DIFF
--- a/docs/HTTP_API.yml
+++ b/docs/HTTP_API.yml
@@ -1,30 +1,16 @@
----
 swagger: "2.0"
-
-############################################################################
-# Info                                                                     #
-############################################################################
 info:
-  version: "0"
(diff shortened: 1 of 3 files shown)


philipnrmn:
Thanks for this @emanic, if you could rebase this branch I'll merge.

emanic:
@philipnrmn Thank you. I have pulled from latest master and squashed the commits. Let me know if this works for you.

PR #89, opened 2017-04-03
Add standalone datadog plugin

philipnrmn:
This PR adds a plugin for DataDog which does not rely on the datadog-agent to receive or send stats. 

The included [Readme](https://github.com/dcos/dcos-metrics/blob/dff3a6f6f9895841b75859dc840a769bbaa987e2/plugins/datadog-standalone/README.md) contains details on how to use the plugin.

This is useful in situations where you:
1. Don't want the datadog-agent running on every node
1. Want to send stats to datadog directly from each node

### DIFF
--- a/plugins/datadog-standalone/datadog-standalone.go
+++ b/plugins/datadog-standalone/datadog-standalone.go
@@ -1,0 +1,218 @@
+// Copyright 2017 Mesosphere, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"math"
+	"net/http"
+	"strconv"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	plugin "github.com/dcos/dcos-metrics/plugins"
+	"github.com/dcos/dcos-metrics/producers"
+	"github.com/urfave/cli"
+)
+
+var (
+	pluginFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "datadog-key",
+			Usage: "DataDog API Key",
+		},
+	}
+	datadogConnector = func(metrics []producers.MetricsMessage, c *cli.Context) error {
+		if len(metrics) == 0 {
+			log.Error("No messages received from metrics service")
+		} else {
+			log.Info("Transmitting metrics to DataDog")
+			datadogURL := fmt.Sprintf("https://app.datadoghq.com/api/v1/series?api_key=%s", c.String("datadog-key"))
+
+			s := messagesToSeries(metrics)
+			b := new(bytes.Buffer)
+			json.NewEncoder(b).Encode(s)
+			res, err := http.Post(datadogURL, "application/json; charset=utf-8", b)
+			if err != nil {
+				log.Error(err)
+				return nil
+			}
+			defer res.Body.Close()
+			body, err := ioutil.ReadAll(res.Body)
+			if err != nil {
+				log.Fatal(err)
+				return nil
+			}
+			result := DDResult{}
+			err = json.Unmarshal(body, &result)
+			if err != nil {
+				log.Error(err)
+				return nil
+			}
+			if len(result.Errors) > 0 {
+				log.Error("Encountered errors:")
+				for _, err := range result.Errors {
+					log.Error(err)
+				}
+			}
+			if len(result.Warnings) > 0 {
+				log.Warn("Encountered warnings:")
+				for _, wrn := range result.Warnings {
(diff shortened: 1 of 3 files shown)


plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -0,0 +1,218 @@
...
+				datapointTags = append(datapointTags, fmt.Sprintf("%s:%s", key, value))
+			}
+			for name, value := range datapoint.Tags {
+				addDatapointTag(name, value)
+			}
+
+			metric := DDMetric{
+				Metric: datapoint.Name,
+				Points: []DDDataPoint{{float64(t.Unix()), v}},
+				Tags:   append(messageTags, datapointTags...),
```
collinvandyck: We found out the other week that we had some slight naming issues with tag names when we were looking at the Librato results. I believe we had a scenario where we could end up with two tags `containerId` from the messages and then `container.id` from the datapoint tags.  Perhaps we should figure that out as a part of this changeset?

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -0,0 +1,218 @@
...
+				Host:   host,
+			}
+			series.Series = append(series.Series, metric)
+		}
+	}
+
+	return series
+}
+
+func toFloat64(value interface{}) (float64, error) {
```
collinvandyck: Should we push this into the `plugin` package and export it?

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -0,0 +1,218 @@
...
+		return float64(value), nil
+	case int:
+		return float64(value), nil
+	case string:
+		return strconv.ParseFloat(value, 64)
+	}
+	return math.NaN(), fmt.Errorf("Could not convert %+v (%T) to float64", value, value)
+}
+
+func parseTimestamp(timestamp string) (*time.Time, error) {
```
collinvandyck: Likewise, since this logic is now duplicated between two pending plugins, should we also move this into the `plugin` package and export it?

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -83,6 +83,8 @@ var (
 			}
 			if result.Status == "ok" {
 				log.Info("Successfully transmitted metrics")
+			} else if result.Status != nil {
```
collinvandyck: this could be

```
if result.Status == "ok" {
  log.Info("Successfully transmitted metrics")
} else {
  log.Warnf("Expected status to be ok, actually: '%v'", result.Status)
}

Comment (philipnrmn) [inline plugins/datadog-standalone/datadog-standalone.go]: I was trying to avoid warning when status is absent, which from what I can make out of the [datadog api reference](http://docs.datadoghq.com/api/), it might be. So I feel like an warning will then end 'expected status to be ok, actually nil'. I guess that's not too awful though, what's your feeling?

Comment (collinvandyck) [inline plugins/datadog-standalone/datadog-standalone.go]: Hmm ok yeah that makes sense then. I wasn't aware that the status could be absent.  If you want to check against an absent status in your original changeset you'll need to compare `result.Status != ""` because go strings can't be nil.

File: plugins/datadog-standalone/datadog-standalone_test.go [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+				log.Fatalf("Expected to find tag %s, but it was not present", expectedTag)
+			}
+		}
+
+		if *uptime.Host != "192.168.65.90" {
+			log.Fatalf("Expected to find host 192.168.65.90, found %s", uptime.Host)
+		}
+
+	}))
+	postMetricsToDatadog(server.URL, []producers.MetricsMessage{nodeMetrics})
```
collinvandyck: you probably want a `defer server.Close()` before `postMetricsToDatadog`.

plugins/datadog-standalone/datadog-standalone_test.go
```diff
@@ -0,0 +1,96 @@
...
+		for _, expectedTag := range expectedTags {
+			tagWasFound := false
+			for _, tag := range uptime.Tags {
+				if expectedTag == tag {
+					tagWasFound = true
+					break
+				}
+			}
+			if !tagWasFound {
+				log.Fatalf("Expected to find tag %s, but it was not present", expectedTag)
```
collinvandyck: `t.Fatalf` instead?

plugins/datadog-standalone/datadog-standalone_test.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+
+		if *uptime.Host != "192.168.65.90" {
+			log.Fatalf("Expected to find host 192.168.65.90, found %s", uptime.Host)
```
collinvandyck: `t.Fatalf` instead?

plugins/datadog-standalone/datadog-standalone_test.go
```diff
@@ -0,0 +1,97 @@
...
+			"mesos_id":"378922ca-dd97-4802-a1b2-dde2f42d74ac",
+			"cluster_id":"cc477141-2c93-4b9f-bf05-ec0e163ce2bd",
+			"hostname":"192.168.65.90"
+		}
+	}`
+)
+
+func TestPostMetricsToDatadog(t *testing.T) {
+	nodeMetrics := producers.MetricsMessage{}
+	json.Unmarshal([]byte(nodeMetricsJSON), &nodeMetrics)
```
collinvandyck: `json.Unmarshal` returns an error

collinvandyck: Other than that, I think this is looking really good.

collinvandyck:
🎉

PR #92, opened 2017-04-06
Changed test to check for monotonically increasing cpu time values.

collinvandyck:
This fixes an issue where the test was checking that the cpu ticks had to always be increasing between samplings.  Because the counter is monotonically increasing is it understood that this value may not change between samplings.  Therefore, the test should allow values that are equal to each other.


philipnrmn:
Just needs a rebase 👍

collinvandyck:
Rebased.

PR #93, opened 2017-04-09
Removed unreachable statement (caught by `go vet`)

collinvandyck:
I didn't actually know whether or not this would compile -- but it does!


philipnrmn:
Interesting.

PR #94, opened 2017-04-26
Fix store cache expiry, fixing stale data bug

philipnrmn:
There was a nasty bug in the metrics package where the cache expiry period was set to 189 years instead of two minutes. 

This PR:-
 - [x] fixes the expiry period
 - [x] adds a test to ensure that this does not occur again
 - [x] corrects the values in the configuration yaml

### DIFF
--- a/dcos-metrics.go
+++ b/dcos-metrics.go
@@ -18,7 +18,6 @@ import (
 	"os"
 	"reflect"
 	"strings"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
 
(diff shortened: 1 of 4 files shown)


dcos-metrics.go
```diff
@@ -58,7 +57,12 @@ func main() {
 	if producerIsConfigured("http", cfg) {
 		log.Info("HTTP producer enabled")
 		cfg.Producers.HTTPProducerConfig.DCOSRole = cfg.DCOSRole
-		cfg.Producers.HTTPProducerConfig.CacheExpiry = time.Duration(cfg.Collector.MesosAgent.PollPeriod) * time.Minute * 2
+
+		minCacheExpiry := cfg.Collector.MesosAgent.PollPeriod * 2
+		if cfg.Producers.HTTPProducerConfig.CacheExpiry < minCacheExpiry {
```
collinvandyck: I'm not sure I really understand this. If the bug was that the cache expiry was set too large, then why are we checking it against a min value?

philipnrmn: I think the [original code](https://github.com/dcos/dcos-metrics/blob/master/dcos-metrics.go#L61) exists to prevent metrics being removed from the cache prematurely (eg if PollPeriod was 1m and CacheExpiry was 30s, metrics would not always be available for a long-running service). 

This was also kind of crappy because a longer cache expiry in the configuration file would be overridden silently. 

We now fix the erroneous multiplication by `time.Minute`, honour the user's configuration if it's sensible and warn them if it's not.

collinvandyck: Ah, ok I understand now. Thanks.

PR #95, opened 2017-05-05
Fix NaN errors in traces

philipnrmn:
At present, we're doing something pretty evil: coercing data into structs by marshalling it to JSON (which uses `reflect` internally 😱) and then unmarshalling it again to `producer.MetricsMessage`:
https://github.com/dcos/dcos-metrics/blob/master/collectors/framework/record.go#L106-L116

This is choking on datapoints which have `NaN` values, which is, in a word, [unfortunate](https://github.com/golang/go/issues/3480#issuecomment-66066048).

### DIFF
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -17,7 +17,9 @@ package framework
 import (
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"io"
+	"math"
 	"net"
 	"time"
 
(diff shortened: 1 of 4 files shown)


philipnrmn:
The approach I took in the end is simply updating `NaN` value datapoints to `"NaN"`. 

Before:
screen shot 2017-05-06 at 8 58 27 pm

After:
screen shot 2017-05-06 at 8 58 37 pm

collectors/framework/framework.go
```diff
@@ -220,6 +222,32 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 		return pmm, err
 	}
 
+	decodedDps, ok := dps.([]interface{})
```
darkonie: would it be possible to type assert to `[]*goavro.Record` ?

collectors/framework/framework.go
```diff
@@ -220,6 +222,32 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
...
+	for _, dp := range decodedDps {
+		dpr, ok := dp.(*goavro.Record)
+		if !ok {
+			fwColLog.Debugf("Bad datapoint record encountered: %v", dp)
+			continue
+		}
+		value, err := dpr.Get(metricsValueKey)
+		if err != nil {
+			fwColLog.Debugf("Datum without value encountered: %v", dpr)
+			continue
```
darkonie: is it ok to have partial result?

collinvandyck: @darkonie I had this same concern but if you look at the context of the code, partial results are returned for most if not all errors (outside of the context of this changeset)

collectors/framework/framework.go
collinvandyck: @darkonie doesn't look that way. I tried it out and i think that type assertion failed (the tests failed at least).  guessing the slice is literally a slice of interface{} to begin with.

PR #96, opened 2017-05-11
test-PR for DC/OS Branch 1.9.1

collinvandyck:
This PR is meant to ensure that the CI tests should pass against this branch that is to be included into the DC/OS 1.9.1 release.


philipnrmn:
the 1.10 dcos-metrics state is compatible with 1.9.x, so I don't think we need this branch any longer.

philipnrmn:
We did still need this branch. It was making the 1.9.1 integration tests pass.

PR #97, opened 2017-05-24
Handle errors gracefully

philipnrmn:
This PR prevents the DataDog plugin crashing when a bad result is returned from the DataDog API. 

For example, when a very long tag is sent to DataDog, the API responds with the text 'Payload too big', instead of JSON. Previously the plugin would have crashed with a JSON parse error. It now logs an error and continues to run.

### DIFF
--- a/plugins/datadog-standalone/datadog-standalone.go
+++ b/plugins/datadog-standalone/datadog-standalone.go
@@ -88,7 +88,8 @@ func datadogConnector(metrics []producers.MetricsMessage, c *cli.Context) error 
 
 	result, err := postMetricsToDatadog(datadogURL, metrics)
 	if err != nil {
-		return err
+		log.Errorf("Unexpected error while processing DataDog response: %v", err)
+		return nil
 	}
 
 	if len(result.Errors) > 0 {
@@ -134,6 +135,7 @@ func postMetricsToDatadog(datadogURL string, metrics []producers.MetricsMessage)
 	result := DDResult{}
 	err = json.Unmarshal(body, &result)
 	if err != nil {
+		log.Warnf("Unreadable DataDog output: %s", body)
 		return nil, err
 	}
 	return &result, nil
--- a/plugins/datadog-standalone/datadog-standalone.go
+++ b/plugins/datadog-standalone/datadog-standalone.go
@@ -88,7 +88,7 @@ func datadogConnector(metrics []producers.MetricsMessage, c *cli.Context) error 
 
 	result, err := postMetricsToDatadog(datadogURL, metrics)
 	if err != nil {
-		log.Errorf("Unexpected error while processing DataDog response: %v", err)
+		log.Errorf("Unexpected error while processing DataDog response: %s", err)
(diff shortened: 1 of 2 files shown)


plugins/plugin.go
```diff
@@ -141,9 +141,12 @@ func (p *Plugin) StartPlugin() error {
 // Metrics polls the DC/OS components and returns a slice of
 // producers.MetricsMessage.
 func (p *Plugin) Metrics() ([]producers.MetricsMessage, error) {
-	p.Log.Info("Getting metrics from metrics service")
 	metricsMessages := []producers.MetricsMessage{}
+	if len(p.AuthToken) == 0 {
+		return metricsMessages, errors.New("Auth token must be set, use --auth-token <token>")
```
collinvandyck: Probably can just return nil for the first return value here.  Also, maybe move this check above the line

`metricsMessages := []producers.MetricsMessage{}`

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -88,7 +88,8 @@ func datadogConnector(metrics []producers.MetricsMessage, c *cli.Context) error
 
 	result, err := postMetricsToDatadog(datadogURL, metrics)
 	if err != nil {
-		return err
+		log.Errorf("Unexpected error while processing DataDog response: %v", err)
+		return nil
```
collinvandyck: +1 to returning nil here.  It's a little unfortunate that if the connector func returns a non-nil err it causes the process to exit.  A future pull request should probably address that because it's janky to just return nil here if an error happens.

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -88,7 +88,8 @@ func datadogConnector(metrics []producers.MetricsMessage, c *cli.Context) error
 
 	result, err := postMetricsToDatadog(datadogURL, metrics)
 	if err != nil {
-		return err
+		log.Errorf("Unexpected error while processing DataDog response: %v", err)
```
collinvandyck: Should probably use `%s` for err here because `error` implements `fmt.Stringer`

plugins/datadog-standalone/datadog-standalone.go
```diff
@@ -134,6 +135,7 @@ func postMetricsToDatadog(datadogURL string, metrics []producers.MetricsMessage)
 	result := DDResult{}
 	err = json.Unmarshal(body, &result)
 	if err != nil {
+		log.Warnf("Unreadable DataDog output: %s", body)
```
collinvandyck: I think it would be good to also include the err here in the log message.

```
log.Warnf("Could not unmarshal datadog response %s: %s", body, err)
```

plugins/datadog-standalone/datadog-standalone.go
philipnrmn: Yeah, this seemed wrong to me too. I'll queue that up as a tech-debt fix.

PR #98, opened 2017-05-30
Drop long labels in mesos agent

philipnrmn:
This PR introduces a check on the length of label values from Mesos. Any label with a value exceeding 128 characters is dropped with a warning. 

Mesos task labels can be extremely long. This becomes a problem with DC/OS packages where labels are used to carry JSON metadata in base64 format. Each label is transformed to a tag per datapoint, causing unnecessary data inflation.

This became urgent when it increased DataDog metrics payloads above 8kb and caused the API to return errors.

### DIFF
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -154,7 +154,7 @@ func getLabelsByContainerID(containerID string, frameworks []frameworkInfo, log 
 				log.Debugf("ContainerID %v for executor %v is a match, adding labels", containerID, executor)
 				for _, pair := range executor.Labels {
 					if len(pair.Value) > maxLabelLength {
-						log.Warnf("Label %s is longer than %d chars; discarding label", maxLabelLength, pair.Key)
+						log.Warnf("Label %s is longer than %d chars; discarding label", pair.Key, maxLabelLength)
 						log.Debugf("Discarded label value: %s", pair.Value)
 						continue
 					}
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -27,7 +27,8 @@ import (
 
 const (
 	// HTTPTIMEOUT defines the maximum duration for all requests
-	HTTPTIMEOUT = 2 * time.Second
+	HTTPTIMEOUT    = 2 * time.Second
+	maxLabelLength = 128
 )
 
 // Collector defines the collector type for Mesos agent. It is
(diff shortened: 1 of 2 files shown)


collectors/mesos/agent/agent.go
```diff
@@ -146,14 +146,19 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
...
-		log.Debugf("Attemping to add labels to %v framework", framework)
+		log.Debugf("Attempting to add labels to %v framework", framework)
 		for _, executor := range framework.Executors {
 			log.Debugf("Found executor %v for framework %v", framework, executor)
 			if executor.Container == containerID {
 				log.Debugf("ContainerID %v for executor %v is a match, adding labels", containerID, executor)
 				for _, pair := range executor.Labels {
-					log.Debugf("Adding label for containerID %v: %v = %+v", containerID, pair.Key, pair.Value)
-					labels[pair.Key] = pair.Value
+					if len(pair.Value) > 128 {
```
collinvandyck: since we use `128` twice here, let's move that to a `const` at the top.

collectors/mesos/agent/agent.go
```diff
@@ -146,14 +146,19 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
...
+						log.Warnf("Label %s is longer than 128 chars; discarding label", pair.Key)
+						log.Debugf("Discarded label value: %s", pair.Value)
```
collinvandyck: Instead of an if/else the usual way this would appear in the wild would be

```
if len(pair.Value) > 128 {
  log.warn(...)
  continue
}
labels[pair.Key] = pair.Value
```

collectors/mesos/agent/agent.go
```diff
@@ -152,8 +153,8 @@ func getLabelsByContainerID(containerID string, frameworks []frameworkInfo, log
 			if executor.Container == containerID {
 				log.Debugf("ContainerID %v for executor %v is a match, adding labels", containerID, executor)
 				for _, pair := range executor.Labels {
-					if len(pair.Value) > 128 {
-						log.Warnf("Label %s is longer than 128 chars; discarding label", pair.Key)
+					if len(pair.Value) > maxLabelLength {
+						log.Warnf("Label %s is longer than %d chars; discarding label", maxLabelLength, pair.Key)
```
collinvandyck: I think you have your arguments mixed up here.

collectors/mesos/agent/agent.go
```diff
@@ -146,12 +147,17 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
...
-		log.Debugf("Attemping to add labels to %v framework", framework)
+		log.Debugf("Attempting to add labels to %v framework", framework)
 		for _, executor := range framework.Executors {
 			log.Debugf("Found executor %v for framework %v", framework, executor)
 			if executor.Container == containerID {
 				log.Debugf("ContainerID %v for executor %v is a match, adding labels", containerID, executor)
 				for _, pair := range executor.Labels {
+					if len(pair.Value) > maxLabelLength {
+						log.Warnf("Label %s is longer than %d chars; discarding label", maxLabelLength, pair.Key)
+						log.Debugf("Discarded label value: %s", pair.Value)
```
collinvandyck: Curious, do we need the Debugf statement anymore?

philipnrmn: I actually think it's really useful to see the actual label that caused the warning when you're debugging. I know it's a little funky to warn and debug one after the other. I guess the user could follow the other debug traces to identify the issue if they had to. What's your feeling?

collinvandyck: Ah, ok, I didn't see initially that the difference between the two was that the WARN was using the  key and the DEBUG was using the value. +++

PR #99, opened 2017-06-18
Use configurable IAM in preference to JWT token in plugins

philipnrmn:
The `--auth-token` parameter previously used by plugins was not a good strategy. If the token expired (in DC/OS configurations which use JWTs, they regularly last five days), the plugin would stop working simultaneously on every node and report errors. The only way to make it work again was to refresh the configuration and restart it. 

With this PR, plugins take a `--config` flag similar to the `--config` flag passed to the dcos-metrics collector, which points to a YAML file with IAM-related information. The plugin loads the file in and uses its contents to authenticate the HTTP client which makes requests against the dcos-metrics API.

### DIFF
--- a/plugins/option.go
+++ b/plugins/option.go
@@ -16,7 +16,6 @@ package plugin
 
 import (
 	"errors"
-	"strconv"
 
 	"github.com/dcos/dcos-metrics/producers"
 	"github.com/urfave/cli"
@@ -56,7 +55,7 @@ func PollingInterval(i int) Option {
 func MetricsProtocol(proto string) Option {
 	return func(p *Plugin) error {
 		if proto == "http" || proto == "https" {
-			p.MetricsProto = proto
+			p.MetricsScheme = proto
 			return nil
 		}
 		return errBadProto
@@ -77,7 +76,7 @@ func MetricsHost(h string) Option {
 func MetricsPort(port int) Option {
 	return func(p *Plugin) error {
 		if port <= 65535 {
-			p.MetricsPort = strconv.Itoa(port)
+			p.MetricsPort = port
 			return nil
 		}
 		return errBadPort
--- a/plugins/option.go
+++ b/plugins/option.go
@@ -84,18 +84,6 @@ func MetricsPort(port int) Option {
 	}
 }
 
-// MetricsAuthToken allows the plugin to set a custom auth token for the
-// Authorization header for the http request that is used to gather metrics.
-func MetricsAuthToken(t string) Option {
-	return func(p *Plugin) error {
-		if len(t) != 0 {
-			p.AuthToken = t
-			return nil
-		}
-		return errBadToken
-	}
-}
-
 // ConnectorFunc is what the plugin framework will call once it has gathered
 // metrics. It is expected that this function will convert these messages to
 // a 3rd party format and then send the metrics to that service.
--- a/plugins/plugin.go
+++ b/plugins/plugin.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -41,7 +42,7 @@ type Plugin struct {
 	Endpoints         []string
 	Role              string
 	PollingInterval   int
-	MetricsPort       string
+	MetricsPort       int
 	MetricsScheme     string
 	MetricsHost       string
 	Log               *logrus.Entry
@@ -62,7 +63,7 @@ func New(options ...Option) (*Plugin, error) {
 		PollingInterval: 10,
 		MetricsScheme:   "http",
 		MetricsHost:     "localhost",
-		MetricsPort:     "61001",
+		MetricsPort:     61001,
 	}
 
 	newPlugin.App = cli.NewApp()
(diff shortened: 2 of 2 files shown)


plugins/plugin.go
```diff
@@ -51,12 +52,13 @@ var version = "UNSET"
...
 func New(options ...Option) (*Plugin, error) {
 	newPlugin := &Plugin{
-		Name:            "default",
-		Role:            "",
-		PollingInterval: 10,
-		MetricsProto:    "http",
-		MetricsHost:     "localhost",
-		MetricsPort:     "61001",
+		Name:              "default",
+		Role:              "",
```
darkonie: by default a string is empty, no need to explicitly initialize it here

plugins/plugin.go
```diff
@@ -51,12 +52,13 @@ var version = "UNSET"
...
+		PollingInterval:   10,
+		MetricsProto:      "http",
+		MetricsHost:       "localhost",
+		MetricsPort:       "61001",
```
darkonie: wonder why is port a string? probably make it int for proper validation

plugins/plugin.go
```diff
@@ -51,12 +52,13 @@ var version = "UNSET"
...
-		Name:            "default",
-		Role:            "",
-		PollingInterval: 10,
-		MetricsProto:    "http",
-		MetricsHost:     "localhost",
-		MetricsPort:     "61001",
+		Name:              "default",
+		Role:              "",
+		PollingInterval:   10,
+		MetricsProto:      "http",
```
darkonie: `MetricsScheme` ?

plugins/plugin.go
```diff
@@ -59,6 +63,8 @@ func New(options ...Option) (*Plugin, error) {
 		MetricsHost:       "localhost",
 		MetricsPort:       "61001",
 		ConfigPath:        "",
+		IAMConfigPath:     "",
+		CACertificatePath: "",
```
darkonie: same here, no need to initialize empty string. Or maybe i am not quite understand why are you doing it?

plugins/plugin.go
```diff
@@ -256,3 +262,19 @@ func makeMetricsRequest(request *http.Request) (producers.MetricsMessage, error)
...
+// loadConfig loads the CACertPath and IAMConfig from the specified yaml file
+// into the corresponding Plugin struct fields
+func (p *Plugin) loadConfig() error {
+	p.Log.Info("Loading optional authentication configuration")
+	fileByte, err := ioutil.ReadFile(p.ConfigPath)
+	if err != nil {
+		return err
+	}
+
+	if err = yaml.Unmarshal(fileByte, &p); err != nil {
```
darkonie: is not `p` a reference already?

plugins/plugin.go
```diff
@@ -45,6 +46,7 @@ type Plugin struct {
 	MetricsHost       string
 	Log               *logrus.Entry
 	ConnectorFunc     func([]producers.MetricsMessage, *cli.Context) error
+	Client            *http.Client
```
darkonie: i would not export `http.Client` you are not going to use it from another package, if you do, something is not right and needs a refactor. Maybe you can explain why do you want to export it?

plugins/plugin.go
philipnrmn: no you're right, I shouldn't initialize these.

plugins/plugin.go
philipnrmn: I don't want to export it, but I wanted not to initialize the Client repeatedly and the only way I could figure out to share context was to hold a reference to the client on the Plugin struct.

philipnrmn:
I also realised that we should use https when the IAM config is present, otherwise this will not work in strict mode.

plugins/plugin.go
philipnrmn: How would you suggest I approach this?

plugins/plugin.go
philipnrmn: it's because `net.JoinHostPort` expects port to be given as a string.

plugins/plugin.go
darkonie: you can make a single instance of http.Client and pass a reference to it to every function it needs.
`MetricsHost` can have unexported field `client`. But this is up to you i think it is more like a nit, not a deal breaker. If you feel like you need to export Client for now, it ok

plugins/plugin.go
darkonie: you can easily convert int to string. Th reason why i'd love to see it as int, is because it is easier to validate. Next PR might be adding some validation login and converting string to int just for validation looks gross.

plugins/option.go
```diff
@@ -77,25 +76,13 @@ func MetricsHost(h string) Option {
 func MetricsPort(port int) Option {
 	return func(p *Plugin) error {
 		if port <= 65535 {
-			p.MetricsPort = strconv.Itoa(port)
+			p.MetricsPort = port
```
darkonie: i'd probably leave it an int and call `strconv.Itoa` rather then have a port as a string

plugins/plugin.go
```diff
@@ -141,32 +145,38 @@ func (p *Plugin) StartPlugin() error {
...
 func (p *Plugin) Metrics() ([]producers.MetricsMessage, error) {
-	if len(p.AuthToken) == 0 {
-		return nil, errors.New("Auth token must be set, use --auth-token <token>")
-	}
 
 	metricsMessages := []producers.MetricsMessage{}
+
+	// The first time Metrics() is called, the plugin client
+	// should be initialised
+	if p.Client == nil {
```
darkonie: why not make `p.Client` as a required argument to [`New` constructor?](https://github.com/dcos/dcos-metrics/blob/e4ae35b8e2bd12d208eb9caeea8c0315fdc72437/plugins/plugin.go#L60) you will be 100% sure the http client is initialized that way

PR #100, opened 2017-06-24
Accept sequential statsd input on app metrics channel

philipnrmn:
The design of the metrics collector expected batched statsd input. When input was received, it was given a name constructed from the type of input plus the container ID, for example `dcos.metrics.app.ba5598ef-9cb7-47a1-8f37-1a7e8bf09e84`. This was stored in memory and retrieved by name when the user hit the appropriate endpoint (in this case `/metrics/v0/containers/ba5598ef-9cb7-47a1-8f37-1a7e8bf09e84/app`)

Unfortunately, most applications which use statsd don't batch their inputs, but emit them one-by-one or in chunks, as is convenient. This was causing distinct metrics to be overwritten repeatedly, since they shared the same ID. When the user hit the endpoint, only the most recent received datum or data was served. 

This PR inserts each incoming datapoint into the store using a fully qualified name made up of the input type and container ID as before, but also the name of the datapoint, eg `dcos.metrics.app.ba5598ef-9cb7-47a1-8f37-1a7e8bf09e84.kafka.network.RequestMetrics.LocalTimeMs.min`. The datapoints are combined when the user hits the appropriate endpoint.

### DIFF
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -256,6 +256,23 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 		return pmm, err
 	}
 
+	// Convert message-level labels to datapoint-level tags
+	labels := pmm.Dimensions.Labels
+	if labels != nil {
+		for n, d := range pmm.Datapoints {
+			tt := make(map[string]string)
+			// TODO(philip) account for tag/label collision
+			for k, v := range d.Tags {
+				tt[k] = v
+			}
+			for k, v := range labels {
+				tt[k] = v
+			}
+			pmm.Datapoints[n].Tags = tt
+		}
+	}
+	pmm.Dimensions.Labels = nil
+
 	return pmm, nil
 }
 
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -257,19 +257,16 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 	}
 
 	// Convert message-level labels to datapoint-level tags
-	labels := pmm.Dimensions.Labels
-	if labels != nil {
-		for i, datapoint := range pmm.Datapoints {
-			tt := make(map[string]string)
-			// TODO(philip) account for tag/label collision
-			for k, v := range datapoint.Tags {
-				tt[k] = v
-			}
-			for k, v := range labels {
-				tt[k] = v
-			}
-			pmm.Datapoints[i].Tags = tt
+	for i, datapoint := range pmm.Datapoints {
+		tt := make(map[string]string)
+		// TODO(philip) account for tag/label collision
+		for k, v := range datapoint.Tags {
+			tt[k] = v
 		}
+		for k, v := range pmm.Dimensions.Labels {
+			tt[k] = v
+		}
+		pmm.Datapoints[i].Tags = tt
 	}
 	pmm.Dimensions.Labels = nil
 
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -260,10 +260,10 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 	for i, datapoint := range pmm.Datapoints {
 		tt := make(map[string]string)
 		// TODO(philip) account for tag/label collision
-		for k, v := range datapoint.Tags {
+		for k, v := range pmm.Dimensions.Labels {
 			tt[k] = v
 		}
-		for k, v := range pmm.Dimensions.Labels {
+		for k, v := range datapoint.Tags {
 			tt[k] = v
 		}
 		pmm.Datapoints[i].Tags = tt
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -259,11 +259,13 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 	// Convert message-level labels to datapoint-level tags
 	for i, datapoint := range pmm.Datapoints {
 		tt := make(map[string]string)
-		// TODO(philip) account for tag/label collision
 		for k, v := range pmm.Dimensions.Labels {
 			tt[k] = v
 		}
 		for k, v := range datapoint.Tags {
+			if tt[k] != "" {
+				fwColLog.Warnf("Datapoint label '%s' (%s) overridden to %s", k, tt[k], v)
+			}
 			tt[k] = v
 		}
 		pmm.Datapoints[i].Tags = tt
--- a/collectors/framework/framework.go
+++ b/collectors/framework/framework.go
@@ -259,16 +259,16 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 	// Convert message-level labels to datapoint-level tags
 	labels := pmm.Dimensions.Labels
 	if labels != nil {
-		for n, d := range pmm.Datapoints {
+		for i, datapoint := range pmm.Datapoints {
 			tt := make(map[string]string)
 			// TODO(philip) account for tag/label collision
-			for k, v := range d.Tags {
+			for k, v := range datapoint.Tags {
 				tt[k] = v
 			}
 			for k, v := range labels {
 				tt[k] = v
 			}
-			pmm.Datapoints[n].Tags = tt
+			pmm.Datapoints[i].Tags = tt
 		}
 	}
 	pmm.Dimensions.Labels = nil
--- a/collectors/framework/framework_test.go
+++ b/collectors/framework/framework_test.go
@@ -110,6 +110,7 @@ func TestTransform(t *testing.T) {
 			So(pmm.Dimensions.MesosID, ShouldEqual, "some-mesos-id")
 			So(pmm.Dimensions.ClusterID, ShouldEqual, "some-cluster-id")
 			So(pmm.Dimensions.Hostname, ShouldEqual, "some-hostname")
+			So(pmm.Dimensions.Labels, ShouldEqual, nil)
 		})
 
 		Convey("Should return an error if AvroDatum didn't contain a goavro.Record", func() {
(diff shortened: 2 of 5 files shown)


producers/http/handlers.go
```diff
@@ -98,14 +98,30 @@ func containerAppHandler(p *producerImpl) http.HandlerFunc {
...
+		if len(containerMetrics) == 0 {
 			httpLog.Errorf("/v0/containers/{id}/app - not found in store: %s", key)
 			http.Error(w, "Key not found in store", http.StatusNoContent)
 			return
 		}
 
-		encode(containerMetrics, w)
+		var combinedMetrics producers.MetricsMessage
+		for _, c := range containerMetrics {
+			if _, ok := c.(producers.MetricsMessage); !ok {
```
collinvandyck: It's more standard here to use a type switch:

```
    switch m := m.(type) {
    case producers.MetricsMessage:
       combinedMetrics.Datapoints = append(combinedMetrics.Datapoints, m.Datapoints...)
       combinedMetrics.Dimensions = metric.Dimensions
    default:
      httpLog.Errorf("/v0/containers/{id}/app - unsupported message type")
      http.Error(w, "Got unsupported message type.", http.StatusInternalServerError
      return
    }
```

producers/http/http.go
```diff
@@ -126,6 +126,10 @@ func (p *producerImpl) writeObjectToStore(d producers.Datapoint, m producers.Met
 	}
 	// e.g. dcos.metrics.app.kafka.server.ReplicaFetcherManager.MaxLag
 	qualifiedName := prefix + producers.MetricNamespaceSep + d.Name
+	for k, v := range d.Tags {
+		// e.g. dcos.metrics.node.network.out.errors.interface.eth0
+		qualifiedName = qualifiedName + producers.MetricNamespaceSep + k + producers.MetricNamespaceSep + v
```
collinvandyck: I'm not sure how I feel about this.  Adding the tag name and value to the metric name opens us up to other metric name collisions.  This can happen when another metric name (without tags) has the same qualified name as another metric after the tags are added to the qualified metric name.

For example `metric=foo.bar.baz` would collide with `metric=foo,tags={bar:baz}`

Especially since we don't control all of the metrics (app metrics) it seems that this will come back to bite us later.  I think it's likely that if we want to do this the right way we will need another approach.

collectors/framework/framework.go
```diff
@@ -256,6 +256,23 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
 		return pmm, err
 	}
 
+	// Convert message-level labels to datapoint-level tags
+	labels := pmm.Dimensions.Labels
+	if labels != nil {
```
collinvandyck: You should not need to check for `nil` here.  Using `range` on a `nil` map should be ok.  This could just be

```
for n, d := range pmm.Dimensions.Labels {
```

collectors/framework/framework.go
```diff
@@ -256,6 +256,23 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
...
+		for n, d := range pmm.Datapoints {
```
collinvandyck: I feel like `n` and `d` here could be named more clearly.

collectors/framework/framework.go
```diff
@@ -256,6 +256,23 @@ func (a *AvroDatum) transform(nodeInfo collectors.NodeInfo) (producers.MetricsMe
...
+			tt := make(map[string]string)
+			// TODO(philip) account for tag/label collision
```
collinvandyck: It seems like this TODO is a requirement.  If there is a collision we should probably log it at the least, and possibly return an error.

collectors/framework/framework_test.go
```diff
@@ -106,10 +106,10 @@ func TestTransform(t *testing.T) {
 			So(pmm.Datapoints[0].Name, ShouldEqual, "some-name")
 			So(pmm.Datapoints[0].Value, ShouldEqual, 42.0)
 			So(pmm.Datapoints[0].Timestamp, ShouldNotEqual, "")
+			So(pmm.Datapoints[0].Tags, ShouldResemble, map[string]string{"some-key": "some-val"})
 			So(pmm.Dimensions.MesosID, ShouldEqual, "some-mesos-id")
 			So(pmm.Dimensions.ClusterID, ShouldEqual, "some-cluster-id")
 			So(pmm.Dimensions.Hostname, ShouldEqual, "some-hostname")
-			So(pmm.Dimensions.Labels, ShouldResemble, map[string]string{"some-key": "some-val"})
```
collinvandyck: I think we should probably test that the `Labels` property is nil.

producers/http/handlers.go
```diff
@@ -182,3 +183,16 @@ func encode(v interface{}, w http.ResponseWriter) {
 		http.Error(w, "Failed to encode value to JSON", http.StatusInternalServerError)
 	}
 }
+
+func combineMessages(mm map[string]interface{}) (producers.MetricsMessage, error) {
```
collinvandyck: I left comments in a previous commit where this lived in a different place.  Would *really* like it if we could prevent things like this happening during a review.

producers/http/handlers.go
```diff
@@ -111,16 +111,10 @@ func containerAppHandler(p *producerImpl) http.HandlerFunc {
 			return
 		}
 
-		var combinedMetrics producers.MetricsMessage
```
collinvandyck: See previous feedback left on commit https://github.com/dcos/dcos-metrics/pull/100/commits/9f44bc0cc9241023c7a39a9bcf62f693c8773b5c which this one overrides.

producers/http/handlers.go
```diff
@@ -27,19 +27,15 @@ import (
 
 func nodeHandler(p *producerImpl) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		var am []interface{}
 		nodeMetrics, err := p.store.GetByRegex(producers.NodeMetricPrefix + ".*")
```
collinvandyck: I know this wasn't part of your changeset, but we should really be escaping any constant strings here that go into `store.GetByRegex` here and elsewhere.

producers/http/handlers.go
```diff
@@ -89,9 +89,12 @@ func containerHandler(p *producerImpl) http.HandlerFunc {
...
 		}
 
-		httpLog.Debugf("Encoding container metrics:\n%+v", containerMetrics)
-
-		encode(containerMetrics, w)
+		combinedMetrics, err := combineMessages(containerMetrics)
+		if err != nil {
+			httpLog.Errorf("/v0/containers/{id} - %s", err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
```
collinvandyck: Elsewhere after checking for `err` you also do a check on `len(combinedMetrics)` and if `0` return an error.  Was leaving that part out here intentional?

producers/http/handlers.go
```diff
@@ -136,11 +136,10 @@ func containerAppMetricHandler(p *producerImpl) http.HandlerFunc {
 			producers.AppMetricPrefix, cid,
 		}, producers.MetricNamespaceSep)
 
-		appMetrics, ok := p.store.Get(key)
-		if !ok {
-			httpLog.Errorf("/v0/containers/{id}/app/{metric-id} - not found in store: %s", key)
-			http.Error(w, "Key not found in store", http.StatusNoContent)
-			return
+		appMetrics, err := p.store.GetByRegex(key + ".*" + mid)
```
collinvandyck: Is this regex expression going to handle finding any metrics whose qualified names have tags appended to them?  For example, previously we created the qualified metric name by doing:

```
qualifiedName = qualifiedName + producers.MetricNamespaceSep + k + producers.MetricNamespaceSep + v
```

Which means that the tag name/value would be appended.

The above regex `key + ".*" + mid` seems like it may not match them?  

Ultimately, I feel like we may need to using some other representation for the `Store` which basically maps "qualified name" -> metrics.  We may need a key that is a struct, instead of a string.

producers/http/handlers.go
```diff
@@ -148,18 +148,13 @@ func containerAppMetricHandler(p *producerImpl) http.HandlerFunc {
...
-					Dimensions: appMetrics.(producers.MetricsMessage).Dimensions,
-				}
-				encode(m, w)
-				return
-			}
+		combinedMetrics, err := combineMessages(appMetrics)
+		if err != nil {
+			httpLog.Errorf("/v0/containers/{id}/app/{metric-id} - %s", err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
 		}
```
collinvandyck: Like before, do we need to error if `len(combinedMetrics) == 0` ?

producers/http/handlers.go
```diff
@@ -58,13 +58,18 @@ func containersHandler(p *producerImpl) http.HandlerFunc {
 			return
 		}
 
+		// There will be multiple messages per container ID
+		uniqueIDs := make(map[string]bool)
 		for _, c := range containerMetrics {
 			if _, ok := c.(producers.MetricsMessage); !ok {
```
collinvandyck: A type assertion switch statement here would be cleaner

producers/http/handlers.go
philipnrmn: You're right, this is bad logic. For what it's worth, this regex actually does fetch the appropriate keys because `store.GetByRegex` uses `regexp.MatchString`, which is confusingly a search rather than a match.

producers/http/handlers.go
philipnrmn: This was already addressed in a subsequent commit 👍

producers/http/handlers.go
philipnrmn: This was already addressed in a subsequent commit 👍

producers/http/handlers.go
```diff
@@ -115,34 +141,28 @@ func containerAppMetricHandler(p *producerImpl) http.HandlerFunc {
 		cid := vars["id"]
 		mid := vars["metric-id"]
 		key := strings.Join([]string{
```
collinvandyck: Minor nit -- probably want to use `joinMetricName` here.

philipnrmn:
Fixed up that suggestion!

PR #107, opened 2017-07-01
Fixed d165f05 to avoid missing dispatch and add actually nested support.

klueska:


philipnrmn:
Am I right in thinking this PR should be closed since #108 was merged, @klueska ?

PR #108, opened 2017-07-01
Implement supportsNesting in IsolatorProcess and dispatch from wrapper

mrbrowning:


philipnrmn:
I use the Github squash-and-merge to preserve the history while keeping master relatively clean.

PR #111, opened 2017-07-22
Sort tags

philipnrmn:
_Fixes https://jira.mesosphere.com/browse/DCOS_OSS-1451_
___
This PR ensures that when a datapoint is set in the store, its tags are sorted first. Before this change, tags were added to a datapoint's key by iterating over a map; however, since map iteration in golang is intentionally randomised, two datapoints with identical names and identical tags could end up with different store keys. 

This resulted in multiple datapoints being shown in some circumstances. It disproportionately affected the `/containers/container-id` endpoint since the datapoints there have multiple tags. A sample output showing stale data:
https://gist.github.com/minyk/826e1856a5638e6ccf51d0fcb32c998a

### DIFF
--- a/producers/http/http.go
+++ b/producers/http/http.go
@@ -117,7 +117,8 @@ func (p *producerImpl) writeObjectToStore(d producers.Datapoint, m producers.Met
 	}
 	// e.g. dcos.metrics.app.[ContainerId].kafka.server.ReplicaFetcherManager.MaxLag
 	qualifiedName := joinMetricName(prefix, d.Name)
-	for k, v := range d.Tags {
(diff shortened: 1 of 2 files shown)


producers/http/http.go
```diff
@@ -126,6 +128,22 @@ func (p *producerImpl) writeObjectToStore(d producers.Datapoint, m producers.Met
 	p.store.Set(qualifiedName, newMessage)
 }
 
+// sortTags turns a map[string]string into a slice of key/values, sorted by key.
+func sortTags(tags map[string]string) [][]string {
```
collinvandyck: You could also return a slice of 2-element arrays from this function  (`[][2]string`).

PR #112, opened 2017-08-01
Fix crash when container statistics are not available

philipnrmn:
The mesos `/containers` endpoint returns an array of containers, [as documented here](http://mesos.apache.org/documentation/latest/endpoints/slave/containers/).

It's possible (although undesirable and, arguably, wrong) for the `statistics` member of the container object to be missing, especially in recently-launched jobs. Unfortunately the mesos collector did not guard against this, and when encountering such a circumstance the whole dcos-metrics-agent process could segfault and crash.

Users who make heavy use of frameworks like Chronos and Metronome could see metrics crash occasionally as a result. 

This PR adds a guard for such cases; container members without statistics will result in a warning and no metrics message for the affected container is sent.

### DIFF
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -102,9 +102,14 @@ func (c *Collector) metricsMessages() (out []producers.MetricsMessage) {
 	t := time.Unix(c.timestamp, 0)
 
 	for _, cm := range c.containerMetrics {
+		datapoints, err := c.createContainerDatapoints(cm)
+		if err != nil {
+			c.log.Warnf("Could not create datapoints for container ID %s: %s", cm.ContainerID, err)
+			continue
+		}
 		msg = producers.MetricsMessage{
 			Name:       producers.ContainerMetricPrefix,
-			Datapoints: c.createContainerDatapoints(cm),
+			Datapoints: datapoints,
 			Timestamp:  t.UTC().Unix(),
 		}
 
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -103,8 +103,8 @@ func (c *Collector) metricsMessages() (out []producers.MetricsMessage) {
 
 	for _, cm := range c.containerMetrics {
 		datapoints, err := c.createContainerDatapoints(cm)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


collectors/mesos/agent/agent.go
```diff
@@ -102,9 +102,14 @@ func (c *Collector) metricsMessages() (out []producers.MetricsMessage) {
 	t := time.Unix(c.timestamp, 0)
 
 	for _, cm := range c.containerMetrics {
+		datapoints, err := c.createContainerDatapoints(cm)
+		if err != nil {
+			c.log.Warnf("Could not create datapoints for container ID %s: %s", cm.ContainerID, err)
```
collinvandyck: Note, you include the container ID in the error message that could come from this call, so may be redundant to include it here as well.

collectors/mesos/agent/agent_test.go
```diff
@@ -273,7 +303,7 @@ func TestBuildDatapoints(t *testing.T) {
 				}
 
 				for _, container := range thisContainerMetrics {
-					result := coll.createContainerDatapoints(container)
+					result, _ := coll.createContainerDatapoints(container)
```
collinvandyck: Would be nice to not ignore an error in the test, but I understand why this is the way it is.

collinvandyck:
I'd also suggest that we may create a sentinel error (something defined at the package level) for this use case.  

```
if container.Statistics == nil {
  return ErrNoStatistics
}
```

The reason is that this is less of an error, and more of an interesting warning. This would allow the client to discriminate handling behavior between "we should log this" and "this is actually an error".  

I'm not saying that this is the best way to go about it, but it seemed like something that we may consider.

philipnrmn:
Hi @collinvandyck , I added a sentinel error to `getContainerDatapoints` - is that what you had in mind?

collinvandyck:
@philipnrmn yeah, except that i was envisioning a package level variable like: 

```
var ErrNoStatistics = errors.New("No statistics found")
```

That could then be passed back from that method.  The caller in this case I believes knows the container for which the error was generated and could then log in based on whether or not `err == ErrNoStatistics`.

philipnrmn:
OK, I updated that last commit to reflect that. Mind giving this one last pass to check you're happy with it @collinvandyck ?

PR #113, opened 2017-08-12
Fall back to HTTP on mesos state fetch error

philipnrmn:
At present, dcos-metrics will poll using https whenever an IAM option is supplied in its config parameters. Otherwise it will use HTTP. This is normally fine, because Mesos will serve a 307 redirect to HTTP when the SSL_DOWNGRADE is enabled, and dcos-metrics respects the redirect, 

However, it's possible for Mesos to get itself into a state where the SSL handshake fails. In this case, metrics crashes with the message "Get https://leader.mesos:5050/state: EOF". 

In this PR we introduce a fallback - if any error occurs during an attempt to get state with HTTPS, we fall back to HTTP instead. 

ping @amitaekbote

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -115,13 +116,18 @@ func (c *Config) getNodeInfo(attemptSSL bool) error {
 	}
 
 	// If there is no available certificate, immediately drop back to HTTP
-	var useSSL = attemptSSL && len(c.IAMConfigPath) > 0
+	useSSL := attemptSSL && len(c.IAMConfigPath) > 0
 	// Create a new DC/OS nodeutil instance
-	var stateURL = "http://leader.mesos:5050/state"
-	if useSSL {
-		stateURL = "https://leader.mesos:5050/state"
+	stateURL := url.URL{
+		Scheme: "http",
+		Host:   "leader.mesos:5050",
+		Path:   "/state",
 	}
-	node, err := nodeutil.NewNodeInfo(client, c.DCOSRole, nodeutil.OptionMesosStateURL(stateURL))
+	if useSSL {
+		stateURL.Scheme = "https"
+	}
+
+	node, err := nodeutil.NewNodeInfo(client, c.DCOSRole, nodeutil.OptionMesosStateURL(stateURL.String()))
 	if err != nil {
 		return fmt.Errorf("error: could not get nodeInfo: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


philipnrmn:
I could not figure out how to replicate the SSL handshake error from mesos to test this. Suggestions very welcome!

config.go
```diff
@@ -135,6 +137,11 @@ func (c *Config) getNodeInfo() error {
 	// Get Mesos master/agent ID
 	mid, err := node.MesosID(nil)
 	if err != nil {
+		// It's possible that we encountered an SSL error
+		if useSSL {
+			log.Warnf("Received an error when attempting to get Mesos ID: %q; falling back to HTTP", err)
```
collinvandyck: I think we should be using `%s` instead of `%q` here.

config.go
```diff
@@ -107,16 +107,18 @@ func (c *Config) loadConfig() error {
...
 		return err
 	}
 
+	// If there is no available certificate, immediately drop back to HTTP
+	var useSSL = attemptSSL && len(c.IAMConfigPath) > 0
 	// Create a new DC/OS nodeutil instance
 	var stateURL = "http://leader.mesos:5050/state"
-	if len(c.IAMConfigPath) > 0 {
+	if useSSL {
 		stateURL = "https://leader.mesos:5050/state"
```
collinvandyck: I know that this is not part of this changeset, but it might be a good opportunity to deduplicate literals here.

```
stateURL := url.URL {
  Scheme: "http",
  Host: "leader.mesos:5050",
  Path: "/state",
}

if useSSL { 
  stateURL.Scheme = "https"
}

```
And then convert the URL to string using `stateURL.String()` below.

config.go
collinvandyck: Oh, nevermind, I saw that for strings, that this quotes the string. Totally makes sense.

amitaekbote:
Verified that this fix causes metric to work while upgrading from  disabled->permissive mode

jgehrcke:
This seems to be one of the last "beta2 blockers". On the one hand we have a code freeze today, on the other hand this does not appear to be shippable yet. @philipnrmn @klueska it seems like we need to make a call here.

philipnrmn:
@jgehrcke I'm not sure why this is a beta blocker. This simply helps to stop one of our integration tests from failing. 

I will need to make a final 1.10 PR when #114 gets merged, I expect this to go in then.

jgehrcke:
> I'm not sure why this is a beta blocker

It's good if it's not :-) Thanks for clarification. I inferred that from a JIRA ticket that linked me to this PR. That JIRA ticket has fixVersion 1.10, and the `beta2` label. Now I have a hard time remembering which JIRA ticket specifically that was.

philipnrmn:
It was likely https://jira.mesosphere.com/browse/DCOS_OSS-1542; I removed the beta2 label from that ticket.

philipnrmn:
The regression test will come in a separate PR.

PR #114, opened 2017-08-15
Add task ID to /container endpoint

philipnrmn:
This PR adds a task ID to the /containers/<container-id> endpoint in dcos-metrics.

### DIFF
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -165,7 +165,7 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
 	return frameworkInfo{}, false
 }
 
-// getExecutorInfoByExecutorID returns the ExecutorInfo struct given its ID.
+// getExecutorInfoByExecutorID returns the executorInfo struct given its ID.
 func getExecutorInfoByExecutorID(executorID string, executors []executorInfo) (executorInfo, bool) {
 	for _, executor := range executors {
 		if executor.ID == executorID {
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -150,6 +150,16 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
 	return frameworkInfo{}, false
 }
 
+// getExecutorInfoByExecutorID returns the ExecutorInfo struct given its ID.
+func getExecutorInfoByExecutorID(executorID string, executors []executorInfo) (executorInfo, bool) {
+	for _, executor := range executors {
+		if executor.ID == executorID {
+			return executor, true
+		}
+	}
+	return executorInfo{}, false
+}
+
 // getLabelsByContainerID returns a map of labels, as specified by the framework
 // that created the executor. In the case of Marathon, the framework allows the
 // user to specify their own arbitrary labels per application.
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -117,23 +117,23 @@ func (c *Collector) metricsMessages() (out []producers.MetricsMessage) {
 			Timestamp:  t.UTC().Unix(),
 		}
 
-		fi, ok := getFrameworkInfoByFrameworkID(cm.FrameworkID, c.agentState.Frameworks)
-		if !ok {
+		fi := getFrameworkInfoByFrameworkID(cm.FrameworkID, c.agentState.Frameworks)
+		if fi == nil {
 			c.log.Warnf("Did not find FrameworkInfo for framework ID %s, skipping!", fi.ID)
 			continue
 		}
 
-		ei, ok := getExecutorInfoByExecutorID(cm.ExecutorID, fi.Executors)
-		if !ok {
+		ei := getExecutorInfoByExecutorID(cm.ExecutorID, fi.Executors)
+		if ei == nil {
 			c.log.Warnf("Did not find ExecutorInfo for executor ID %s, skipping!", cm.ExecutorID)
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


collectors/mesos/agent/state.go
```diff
@@ -50,22 +50,22 @@ type frameworkInfo struct {
...
-	Tasks     []taskInfo       `json:"tasks,omitempty"`
+	ID        string       `json:"id"`
+	Name      string       `json:"name"`
+	Container string       `json:"container"`
+	Labels    []stateLabel `json:"labels,omitempty"` // labels are optional
+	Tasks     []taskInfo   `json:"tasks,omitempty"`
 }
 
-type executorLabels struct {
+type stateLabel struct {
```
collinvandyck: Would it make sense to just name it `label` instead of `stateLabel`?

collectors/mesos/agent/agent.go
```diff
@@ -150,6 +150,16 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
 	return frameworkInfo{}, false
 }
 
+// getExecutorInfoByExecutorID returns the ExecutorInfo struct given its ID.
```
collinvandyck: s/ExecutorInfo/executorInfo ?

collectors/mesos/agent/agent.go
```diff
@@ -150,6 +150,16 @@ func getFrameworkInfoByFrameworkID(frameworkID string, frameworks []frameworkInf
...
+func getExecutorInfoByExecutorID(executorID string, executors []executorInfo) (executorInfo, bool) {
```
collinvandyck: I think this method could just return a `*executorInfo` since the boolean value only signals whether or not you found it.  Nil/Non-Nil would do the same thing as well as prevent an unnecessary allocation if you could not find the executor for the given ID.

collectors/mesos/agent/agent_test.go
```diff
@@ -393,6 +393,29 @@ func TestGetFrameworkInfoByFrameworkID(t *testing.T) {
 	})
 }
 
+func TestGetExecutorInfoByExecutorID(t *testing.T) {
```
collinvandyck: I like that you included the test case in this commit.

collectors/mesos/agent/agent.go
```diff
@@ -160,6 +160,16 @@ func getExecutorInfoByExecutorID(executorID string, executors []executorInfo) (e
 	return executorInfo{}, false
 }
 
+// getTaskInfoByContainerID returns the TaskInfo struct matching the given cID.
+func getTaskInfoByContainerID(containerID string, tasks []taskInfo) (taskInfo, bool) {
```
collinvandyck: Same comment here.. You can just return a `*taskInfo` instead.

collectors/mesos/agent/agent_test.go
```diff
@@ -349,18 +349,27 @@ func TestTransform(t *testing.T) {
 		if err := json.Unmarshal(mockAgentState, &mac.agentState); err != nil {
 			panic(err)
 		}
-		if err := json.Unmarshal(deficientContainerMetrics, &mac.containerMetrics); err != nil {
-			panic(err)
-		}
-
-		Convey("Should return a []producers.MetricsMessage without errors", func() {
+		Convey("With container metrics", func() {
+			if err := json.Unmarshal(mockContainerMetrics, &mac.containerMetrics); err != nil {
```
collinvandyck: Is `mac` re-initialized for each sub-test `Convey(...)` method?  Is there any possibility that there could be dirty state in `mac.containerMetrics` from test-to-test?

collectors/mesos/agent/agent.go
```diff
@@ -123,6 +123,19 @@ func (c *Collector) metricsMessages() (out []producers.MetricsMessage) {
...
+		if !ok {
+			c.log.Warnf("Did not find ExecutorInfo for executor ID %s, skipping!", cm.ExecutorID)
+			continue
+		}
+
+		ti, ok := getTaskInfoByContainerID(cm.ContainerID, ei.Tasks)
+		if !ok {
+			// This is not a warning because task ID is not guaranteed to be set, eg
+			// a custom executor is a container with no associated task.
+			c.log.Debugf("Did not find TaskInfo for container ID %s, skipping!", cm.ContainerID)
```
collinvandyck: Do you need to `continue` here?  Later on you dereference `ti.ID` and `ti.Name` even if you could not find the task info by container id.

collectors/mesos/agent/state.go
philipnrmn: I actually did this at first, but then backed it out and went with `stateLabel` because I thought `label` was too vague. I'll switch it back if you're happy with `label`.

collinvandyck: I'll leave it up to you, but my inclination would be to name it either `label` or `keyValue` because that's all that's in it.

collectors/mesos/agent/agent.go
collinvandyck: In that case, should we also not skip this `cm` if the executor info could not be found?

philipnrmn: A container should always have an associated executor and framework, it's just the task association that isn't guaranteed. I think it's correct to skip and warn if no associated executor is found, because otherwise we'd be serving an unidentifiable metric.

collectors/mesos/agent/agent_test.go
philipnrmn: Yep, I just tested this by adding to a counter each time it was re-initialized and asserting in each sub `Convey` method that it had incremented.

philipnrmn:
I've fixed up all feedback, now merging optimistically.

PR #115, opened 2017-08-24
DO NOT MERGE: Support Framework Authentication Mesos

ts3ng:
Do not merge in current state this PR is informational only for now and still under works.  There is a bug where if http credentials aren't provided it will not work. Will update PR in a bit.

Case: when you enable Mesos Authentication http://mesos.apache.org/documentation/latest/authentication/.  metrics cannot talk to mesos.  Here are changes to support provide basic authorization headers.


ts3ng:
There is a current bug in the PR where if you do not provide credentials it will break... Will post a fix in a bit.

ts3ng:
closing so this won't be merged.

PR #116, opened 2017-08-27
Draw container IDs in /containers from both container and app metrics

philipnrmn:
The `/containers` endpoint currently works by finding all container metrics in memory and returning the set of their container IDs. 

Unfortunately, it's possible for a container to exist which does not have associated container metrics (see https://issues.apache.org/jira/browse/MESOS-7918 for context). Since the ID for this container is never given in `/containers`, it cannot be discovered via the API. 

This PR updates the logic in the `/containers` handler. It now finds all container *and app* metrics in memory and returns the set of their container IDs.

### DIFF
--- a/producers/http/handlers.go
+++ b/producers/http/handlers.go
@@ -51,7 +51,13 @@ func nodeHandler(p *producerImpl) http.HandlerFunc {
 func containersHandler(p *producerImpl) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		cm := []string{}
-		containerMetrics, err := p.store.GetByRegex(regexp.QuoteMeta(producers.ContainerMetricPrefix) + ".*")
+		containerMetrics, err := p.store.GetByRegex(
+			fmt.Sprintf(
+				"(%s|%s).*",
+				regexp.QuoteMeta(producers.ContainerMetricPrefix),
+				regexp.QuoteMeta(producers.AppMetricPrefix),
+			),
+		)
 		if err != nil {
 			httpLog.Errorf("/v0/containers - %s", err.Error())
 			http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/producers/http/handlers_test.go
+++ b/producers/http/handlers_test.go
@@ -171,18 +171,14 @@ func TestContainersHandler(t *testing.T) {
 			}
 			defer resp.Body.Close()
 
-			got, err := ioutil.ReadAll(resp.Body)
-			if err != nil {
-				panic(err)
-			}
 			// Note that we didn't supply testContainerData in `setup` above.
-			expected, err := json.Marshal([]string{testContainerData.Dimensions.ContainerID})
-			if err != nil {
+			var containerIDs []string
+			if err := json.NewDecoder(resp.Body).Decode(&containerIDs); err != nil {
 				panic(err)
 			}
 
 			So(resp.StatusCode, ShouldEqual, http.StatusOK)
-			So(strings.TrimSpace(string(got)), ShouldEqual, strings.TrimSpace(string(expected)))
+			So(containerIDs, ShouldResemble, []string{testContainerData.Dimensions.ContainerID})
 		})
 	})
 }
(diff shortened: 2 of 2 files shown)


producers/http/handlers_test.go
```diff
@@ -1,5 +1,3 @@
-//+build unit
-
```
philipnrmn: This should not have been removed - I corrected it in a later commit. I didn't squash it in, in case doing so broke the downstream DC/OS PR.

producers/http/handlers_test.go
```diff
@@ -86,9 +86,15 @@ var (
 		},
 		Timestamp: testTime.UTC().Unix(),
 	}
+
+	allTestData = []producers.MetricsMessage{
```
klueska: This variable would be more appropriately called `allTestMessages`

producers/http/handlers.go
```diff
@@ -51,7 +51,13 @@ func nodeHandler(p *producerImpl) http.HandlerFunc {
...
 	return func(w http.ResponseWriter, r *http.Request) {
 		cm := []string{}
-		containerMetrics, err := p.store.GetByRegex(regexp.QuoteMeta(producers.ContainerMetricPrefix) + ".*")
+		containerMetrics, err := p.store.GetByRegex(
+			fmt.Sprintf(
+				"(%s|%s).*",
+				regexp.QuoteMeta(producers.ContainerMetricPrefix),
+				regexp.QuoteMeta(producers.AppMetricPrefix),
+			),
+		)
```
klueska: It's not clear to me exactly how this works. I don't see how we get container metrics for nested containers using this.

collinvandyck: What I understand about this is that because all of the app metrics should have the container ID set in the dimensions, even those container IDs that the mesos agent fails to provide from /containers. This allows the container IDs to be successfully collected.

producers/http/handlers_test.go
```diff
@@ -161,6 +161,27 @@ func TestContainersHandler(t *testing.T) {
...
 		})
+		Convey("Should return container IDs when app metrics, but not container metrics, are present", func() {
+			port := setup([]producers.MetricsMessage{testNodeData, testAppData})
+			resp, err := http.Get(urlBuilder("localhost", port, "/v0/containers"))
+			if err != nil {
+				panic(err)
+			}
+			defer resp.Body.Close()
+
+			got, err := ioutil.ReadAll(resp.Body)
```
collinvandyck: I think I'd prefer this test to readlike:

```
var containerIDs []string
if err := json.NewDecoder(resp.Body).Decode(&containerIDs); err != nil {
    panic(err)
}
So(resp.StatusCode, ShouldEqual, http.StatusOK)
So(containerIDs, ShouldResemble, []string{testContainerData.Dimensions.ContainerID})
```

Assuming that ShouldResemble is something like `reflect.DeepEquals`

Not a huge deal but that would be more readable in my opinion.

producers/http/handlers.go
philipnrmn: We don't get container metrics for nested containers (I don't know of any way to do so at present) - hitting `/metrics/v0/containers/<nested-container-id>` will result in a 204 empty response. What this change does is make sure that both the nested container and its parent have their IDs in `/metrics/v0/containers`.

klueska: So we now have nested container IDs in `/metrics/v0/containers/`, but hitting `/metrics/v0/containers/<nested-container-id>` will return a 204? What did it do before this change? Return a 404 because we didn't even have a reference to the container? This seems strange to me, but if mesos doesn't return any metrics about nested containers, I guess there's not much we can do about it -- and returning 202 No Content, seems like the proper response.

So I guess the real point of this change is so we can now hit `/metrics/v0/containers/<nested-container-id>/app` and get some app metrics for this container. Before this change, this wasn't possible because we weren't tracking the nested container IDs under `/metrics/v0/containers/` at all.

Is this an accurate assessment?

philipnrmn: That's almost accurate - it's always been possible to hit `/metrics/v0/containers/<nested-container-id>/app` and get metrics. But the user would have had to go figure out what the nested container's ID was themselves instead of finding it in the `/metrics/v0/containers` endpoint.

Before this PR, hitting `/metrics/v0/containers/<nested-container-id>` endpoint would have returned a 404. This PR literally only adds nested container IDs to the container endpoint using data we already had available.

producers/http/handlers_test.go
philipnrmn: I was following the conventions in the test file, but this is way neater.

philipnrmn:
All feedback responded to/fixed up, as far as I know.

PR #118, opened 2017-09-13
Add task metadata to /apps endpoint

philipnrmn:
This PR is the successor to #114 . It adds the same data - task ID and task name - to the 
`/metrics/v0/containers/<container-id>/app` endpoint as #114 added to the `/metrics/v0/containers/<container-id>` endpoint. 

It had to be separated because the two collectors don't share data. Since task IDs are derived from the agent state, adding them to the agent collector was not especially complicated. Adding them to the framework collector was a bit trickier. 

In order to do so, we add a mapping between container IDs and references to task info (`ContainerTaskRels`), held on the agent collector. From `dcos-metrics.go` where both are initiated, we then pass a reference to the `ContainerTaskRels` object into the framework collector. When the framework collector receives a `container_id` tag, it looks up the associated task metadata and adds it to the metrics message.

### DIFF
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -51,9 +51,11 @@ type Collector struct {
 	// https://godoc.org/github.com/sirupsen/logrus#Entry
 	log *logrus.Entry
 
-	metricsChan chan producers.MetricsMessage
-	nodeInfo    collectors.NodeInfo
-	timestamp   int64
+	metricsChan       chan producers.MetricsMessage
+	nodeInfo          collectors.NodeInfo
+	timestamp         int64
+	ContainerTaskRels ContainerTaskRels
+}
 
 // ContainerTaskRels defines the relationship between containers and tasks.
 type ContainerTaskRels struct {
@@ -109,6 +111,7 @@ func New(cfg Collector, nodeInfo collectors.NodeInfo) (Collector, chan producers
 	c.log = logrus.WithFields(logrus.Fields{"collector": "mesos-agent"})
 	c.nodeInfo = nodeInfo
 	c.metricsChan = make(chan producers.MetricsMessage)
+	c.ContainerTaskRels = ContainerTaskRels{}
 	return c, c.metricsChan
 }
 
@@ -135,6 +138,9 @@ func (c *Collector) pollMesosAgent() {
 	if err := c.getAgentState(); err != nil {
 		c.log.Errorf("Failed to get agent state from %s. Error: %s", host, err)
 	}
+
+	c.log.Debug("Mapping containers to tasks")
+	c.ContainerTaskRels.update(c.agentState)
 
 	c.log.Debugf("Fetching container metrics from host %s", host)
 	if err := c.getContainerMetrics(); err != nil {
--- a/collectors/mesos/agent/agent.go
+++ b/collectors/mesos/agent/agent.go
@@ -179,7 +179,7 @@ func getExecutorInfoByExecutorID(executorID string, executors []executorInfo) *e
 }
 
 // getTaskInfoByContainerID returns the TaskInfo struct matching the given cID.
-func getTaskInfoByContainerID(containerID string, tasks []taskInfo) *taskInfo {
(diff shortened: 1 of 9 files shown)


philipnrmn:
Hi @collinvandyck, thanks for all the feedback!

I just wanted to address this bit from your comment above:
>  I wonder if it makes sense to not emit a time series from the framework collector if no taskID or taskName could be found?

Unfortunately that is not possible, because some containers don't have associated tasks. It's not possible to know when dealing with a container ID in isolation whether you expect an associated task. You could simply not return metrics at all if agentState is not populated, but that doesn't fix the problem either - if a freshly launched container emits metrics, we will need to offer them in the API before we know the associated task ID, if one even exists. 

I don't see any way around this unfortunately.

collinvandyck:
@philipnrmn perhaps we should unite the agent and framework polling mechanism so that the agent always polls first and then, once successful, the framework gets polled next, supplying data from the former to the latter?

philipnrmn:
@collinvandyck that is unfortunately also not possible; the framework collector does not poll - it receives data from apps emitting statsd messages over UDP (via the mesos module). I don't believe it is possible to force consistency.

collectors/mesos/agent/agent.go
```diff
@@ -53,6 +54,53 @@ type Collector struct {
...
+
+// update denormalizes the (deeply nested) /state map from the local mesos
+// agent to a list of tasks mapped to container IDs.
+func (ctr *ContainerTaskRels) update(as agentState) {
+	rels := map[string]*TaskInfo{}
+	for _, f := range as.Frameworks {
+		for _, e := range f.Executors {
+			for _, t := range e.Tasks {
+				for _, s := range t.Statuses {
+					rels[s.ContainerStatusInfo.ID.Value] = &t
```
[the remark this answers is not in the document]
philipnrmn: I absolutely follow the logic of what you're saying, but I'm a little puzzled because I can't force this bug to emerge. I am running four identical applications on the same node, each called emitter-N where N is in the range 1-4. 

It seems to me that if this bug exists, the metrics results for all four containers should have the same name and task ID. But their task IDs remain distinct, correct and consistent.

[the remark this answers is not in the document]
collinvandyck: Since the loop goes first over executors, and then tasks, this bug will not appear if there is only ever one task per executor, I think.  Is it possible that an executor can ever have more than one task?

philipnrmn:
Hey @collinvandyck , I addressed all your feedback. Would we be OK to merge?

collectors/mesos/agent/agent.go
```diff
@@ -95,7 +95,10 @@ func (ctr *ContainerTaskRels) update(as agentState) {
 		for _, e := range f.Executors {
 			for _, t := range e.Tasks {
 				for _, s := range t.Statuses {
-					rels[s.ContainerStatusInfo.ID.Value] = &t
+					rels[s.ContainerStatusInfo.ID.Value] = &TaskInfo{
```
collinvandyck: 👍

collinvandyck: 👍

collectors/mesos/agent/agent.go
```diff
@@ -50,17 +51,75 @@ type Collector struct {
...
+	return &ContainerTaskRels{rels: make(map[string]*TaskInfo)}
+}
+
+// Get is a utility method which handles the mutex lock and abstracts the inner
+// map in ContainerTaskRels away. If no task info is available for the supplied
+// containerID, returns nil.
+func (ctr *ContainerTaskRels) Get(containerID string) *TaskInfo {
+	ctr.Lock()
+	defer ctr.Unlock()
+	if ctr.rels != nil {
```
collinvandyck: Do you need this anymore if you're already constructing the type with a non-nil map, here and otherwise?

collinvandyck:
Hi @philipnrmn yeah, good to go. one or two minor comments here an there.

collectors/mesos/agent/agent.go
philipnrmn: No, I guess not. Just disguising my laziness behind a front of belt-and-braces.

PR #119, opened 2017-09-20
Avoid unregistering unknown containers.

rukletsov:
For debug containers, metrics are not emitted and hence the
container is not registered with the `ContainerAssigner`
instance in the `prepare()` call on the `Isolator` interface.
Hence we should not try to unregister the container in `cleanup()`.

See https://github.com/apache/mesos/blob/11de04c1bbf65d0a90576642f8ea9574e05a317a/include/mesos/slave/isolator.hpp#L115-L124


gpaul:
:tada: Thanks @rukletsov

PR #120, opened 2017-09-22
Reinstate golint

philipnrmn:
This PR reinstates golint to the build (skipped in #118), and fixes two instances of the `return nil` issue flagged by the new rule.


philipnrmn:
❤️

PR #122, opened 2017-09-25
Add -u flag to all `go get`s in docs

philipnrmn:
The -u flag pulls in all dependencies, ensuring that libraries like jason and logrus are available when the user comes to run `make`.

Thanks @frankscholten for the report!

### DIFF
--- a/plugins/stdout/README.md
+++ b/plugins/stdout/README.md
@@ -8,7 +8,7 @@ run it in production.
 ## Usage
 
 ### Build this plugin (requires a Golang environment)
-1. `go get github.com/dcos/dcos-metrics`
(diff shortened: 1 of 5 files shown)


plugins/stdout/README.md
```diff
@@ -25,6 +25,7 @@ build
 Upload the plugin to your node via `scp` or similar, then simply run
 
 `./dcos-metrics-stdout-plugin --dcos-role=agent`
+`./dcos-metrics-stdout-plugin --dcos-role=agent`
```
collinvandyck: Is this line duplicated by accident?

philipnrmn:
It was! thanks for the catch :)

PR #124, opened 2017-10-19
Add statsd plugin

philipnrmn:
This PR adds a plugin for statsd, and documentation which covers how to set up a statsd server on DC/OS. 

screen shot 2017-10-19 at 2 10 45 pm

### DIFF
--- a/plugins/statsd/README.md
+++ b/plugins/statsd/README.md
@@ -1,0 +1,86 @@
+# Statsd plugin for DC/OS Metrics
+
+This plugin will forward all data received by dcos-metrics to a statsd
+server, as defined by STATSD_HOST and STATSD_PORT environment variables.
+
+## Architecture
+
+The statsd plugin uses the dcos-metrics plugin API. You run the 'plugin' (a
+slight misnomer; it is in fact a separate binary) on each node in your cluster.
+It polls its local HTTP API for metrics, which it re-transmits to your statsd
+server.
+
+### Limitations
+
+ - statsd has no units, therefore *units are not sent*
+ - statsd has no tags, therefore *dimensions are not sent*
+ - statsd only accepts integer values, therefore *metrics are rounded to the
+   nearest integer*
+ - statsd assigns its own timestamps when metrics are received, therefore
+   *timestamps do not precisely reflect the moment the measurement occurred*
+ - dcos-metrics has no concept of timers or counters, therefore *all datapoints
+   are sent as gauges*
+
+## Usage
+
+### Installing and starting the plugin
+
+1. Download the latest statsd plugin binary from the [releases](https://github.com/dcos/dcos-metrics/releases) page
+1. Upload the binary as `/opt/mesosphere/bin/dcos-metrics-statsd-plugin` on every node in your cluster
+1. Download the [environment file](./systemd/dcos-metrics-statsd.env)
+1. Upload the environment file to `/opt/mesosphere/etc` on every node
+1. On every master node:
+    1. Copy the [master systemd service](./systemd/dcos-metrics-statsd-master.service) file to `/etc/systemd/system`
+    1. Reload the systemd state by running `systemctl daemon-reload`
+    1. Start the systemd service with `systemctl start dcos-metrics-statsd-master`
+1. On every agent node:
+    1. Copy the [agent systemd service](./systemd/dcos-metrics-statsd-agent.service) file to `/etc/systemd/system`
+    1. Reload the systemd state by running `systemctl daemon-reload`
+    1. Start the systemd service with `systemctl start dcos-metrics-statsd-agent`
+
+### Running a statsd server on a DC/OS cluster
+
+If you installed the plugin according to the steps above, you can simply deploy
+the [statsd server](./marathon/statsd-server.json) application with Marathon.
+The simplest way to do that is using the CLI command
+
+`dcos marathon app add statsd-server.json`
+
+You can also deploy it manually with the DC/OS UI.
+
+This will start a Docker container with a statsd server configured and ready to
+go. You can configure your own routing using marathon-haproxy or edge-LB, or
+you can VPN into your cluster using the [dcos tunnel](https://dcos.io/docs/administration/access-node/tunnel/)
+command, after which you can find the Graphite UI at:
+
+http://statsd-server.marathon.l4lb.thisdcos.directory
+
+Please note that this configuration is neither highly available nor persistent;
+you will need to adapt it if you want those attributes.
+
+### Configuring the plugin to use your own statsd server
+
+If you'd prefer to configure your own statsd server (or already have one set
+up), you'll need to pass its address to the plugin. This requires modifying the
+systemd environment file on every node.
+
+1. Download the [environment file](./systemd/dcos-metrics-statsd.env)
+1. Modify the `STATSD_UDP_HOST` and `STATSD_UDP_PORT` EVs to your desired value
+1. Upload the environment file to `/opt/mesosphere/etc` on every node
+1. Reload systemd state by running `systemctl daemon-reload` on every node
+1. Restart the systemd services by running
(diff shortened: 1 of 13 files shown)


philipnrmn:
retest this please

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		log.Fatal(err)
+	}
+
+	log.Fatal(statsdPlugin.StartPlugin())
+}
+
+// statsdConnector is the method called by the plugin every time it retrieves
+// metrics from the API.
+func statsdConnector(metrics []producers.MetricsMessage, c *cli.Context) error {
+	if statsdClient == nil {
```
collinvandyck: I'd probably wrap this initialization in a `sync.Once` since you're mutating global state with a function that something else is calling.

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		log.Infof("Setting up new statsd client %s:%d", statsdHost, statsdPort)
+		statsdClient = statsd.New(statsdHost, statsdPort)
+	}
+
+	if len(metrics) == 0 {
+		log.Info("No messages received from metrics service")
+		return nil
+	}
+	for _, message := range metrics {
+		log.Debugf("Sending %d datapoints to statsd server...")
```
collinvandyck: You're missing `len(message.Datapoints)` at the end of `log.Debugf`

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	client.Send(data, 1)
+	return nil
+}
+
+// convertDatapointToStatsd attempts to convert a datapoint to a statsd format
+// name + value, returning a false ok flag if the conversion failed.
+func convertDatapointToStatsd(datapoint producers.Datapoint) (string, string, bool) {
+	// Value is of type interface{}, hence sprintf
```
collinvandyck: There's no sprintf here -- does this comment belong somewhere else?

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+	val, err := normalize(datapoint.Value)
+	if err != nil {
+		// This is only debug-level because we expect many NaNs in regular usage
+		log.Debugf("Metric %s: %q", datapoint.Name, err)
```
collinvandyck: Should the log message here be something like 
```
log.Debugf("Metric %s failed to convert: %q")
```

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil
+}
+
+// convertDatapointToStatsd attempts to convert a datapoint to a statsd format
+// name + value, returning a false ok flag if the conversion failed.
+func convertDatapointToStatsd(datapoint producers.Datapoint) (string, string, bool) {
+	// Value is of type interface{}, hence sprintf
+	val, err := normalize(datapoint.Value)
+	if err != nil {
+		// This is only debug-level because we expect many NaNs in regular usage
```
collinvandyck: Just curious, why do we expect many NaNs?

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		return "", "", false
+	}
+	return datapoint.Name, fmt.Sprintf("%d|g", val), true
+}
+
+// normalize converts ints, floats and strings to rounded ints
+// It errors on other types and NaNs.
+func normalize(i interface{}) (int, error) {
+	switch v := i.(type) {
+	case int:
```
collinvandyck: Does this catch 32bit ints? I'm not sure if that's something we'd expect or not here, but might be worth looking into.

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		return v, nil
+	case float64:
```
collinvandyck: Same here, do we want to address 32 bit floats?

plugins/statsd/statsd.go
```diff
@@ -0,0 +1,135 @@
...
+		// We need this check here because NaN can be a float. Casting NaN to
+		// int results in a large negative number. Even after you add 0.5.
+		if math.IsNaN(v) {
+			return -1, fmt.Errorf("Could not normalize NaN %q", v)
+		}
+		return int(v + 0.5), nil
+	case string:
+		f, err := strconv.ParseFloat(v, 32)
```
collinvandyck: Why are you specifying a bitSize of 32 instead of 64?

plugins/statsd/statsd_test.go
```diff
@@ -0,0 +1,148 @@
...
+	// This closely follows how the plugin is constructed.
+	app := cli.NewApp()
+	app.Flags = pluginFlags
+
+	Convey("When supplying metrics to the connector", t, func() {
+		app.Action = func(c *cli.Context) {
+			// Inside app.Action, the cli context is appropriate populated
+			err := statsdConnector([]producers.MetricsMessage{fooBarMetric}, c)
+			So(err, ShouldBeNil)
+			metric := <-msgs
```
collinvandyck: Do you want to fetch the second metric as well? `foo.baz`

plugins/statsd/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The statsd plugin uses the dcos-metrics plugin API. You run the 'plugin' (a
+slight misnomer; it is in fact a separate binary) on each node in your cluster.
+It polls its local HTTP API for metrics, which it re-transmits to your statsd
+server.
+
+### Limitations
+
+ - statsd has no units, therefore *units are not sent*
+ - statsd has no tags, therefore *dimensions are not sent*
```
collinvandyck: If we are discarding dimensions does that mean that metrics will be overwritten in statsd, or does it mean that metrics will all be aggregates?

philipnrmn: The metrics are gauges, so the most recent one will overwrite the previous one. The effect is that identical metrics from multiple sources are not useful. This is a statsd limitation.

plugins/statsd/statsd.go
philipnrmn: Null values, valid in JSON, come out as NaN. There are many null values in codahale metrics - metrics which cassandra and kafka and others use - because of percentile calcuations on zero counts.

plugins/statsd/statsd_test.go
philipnrmn: Thanks for pointing this out - it caused me to go and run the test repeatedly, and (unsurprisingly in retrospect) order of delivery is not guaranteed and this test would sometimes flake.

plugins/statsd/statsd.go
```diff
@@ -100,13 +100,13 @@ func emitDatapointsOverStatsd(datapoints []producers.Datapoint, client *statsd.S
...
 // name + value, returning a false ok flag if the conversion failed.
 func convertDatapointToStatsd(datapoint producers.Datapoint) (string, string, bool) {
-	// Value is of type interface{}, hence sprintf
 	val, err := normalize(datapoint.Value)
 	if err != nil {
 		// This is only debug-level because we expect many NaNs in regular usage
 		log.Debugf("Metric %s failed to convert: %q", datapoint.Name, err)
 		return "", "", false
 	}
+	// Value is of type interface{}, hence sprintf
```
collinvandyck: Looks like val here is an int, not an interface{}

plugins/statsd/statsd_test.go
```diff
@@ -122,12 +122,15 @@ func TestStatsdConnector(t *testing.T) {
...
-		"123.54": 124,
+		123:             123,
+		int32(123):      123,
+		int64(123):      123,
+		float32(123.45): 123,
+		float64(123.45): 123,
+		123.54:          124,
+		"123":           123,
+		"123.45":        123,
+		"123.54":        124,
```
collinvandyck: Nice

PR #127, opened 2017-11-01
Prometheus plugin

philipnrmn:
This plugin offers a `/metrics` endpoint on the nominated local port which serves all the content available locally, appropriately tagged. 

screen shot 2017-11-01 at 10 44 02 am


plugins/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,62 @@
...
+		app.Run([]string{"", "-prometheus-port", "8080"})
+	})
+}
+
+// getLocalMetrics fetches text from http://localhost:port/metrics
+func getLocalMetrics(port int) (string, int) {
+	addr := fmt.Sprintf("http://localhost:%d/metrics", port)
+	response, err := http.Get(addr)
+	if err != nil {
+		panic(err)
```
collinvandyck: Seems odd to panic in a test... Do we want to return the error here instead of panicking?  Alternatively, pass `t` into the helper method so you can call `t.Fatal(err)`?

plugins/prometheus/prometheus_test.go
```diff
@@ -42,10 +43,22 @@ func TestPromPlugin(t *testing.T) {
 			return nil
 
 		}
-		app.Run([]string{"", "-prometheus-port", "8080"})
+		app.Run([]string{"", "-prometheus-port", freeport()})
 	})
 }
 
+// freeport listens momentarily on port 0, then closes the connection and
```
collinvandyck: This approach seems like it would make for a flaky test, but I'm not sure what the alternative would be since you need to pass the port as an int to the app.

plugins/prometheus/prometheus.go
```diff
@@ -94,3 +97,22 @@ func promConnector(metrics []producers.MetricsMessage, c *cli.Context) error {
...
+	var buffer bytes.Buffer
+
+	for _, d := range message.Datapoints {
+		name := strings.Replace(d.Name, ".", "_", -1)
+		t, err := time.Parse(time.RFC3339, d.Timestamp)
+		if err != nil {
+			log.Warnf("Encountered bad timestamp, %q: %s", d.Timestamp, err)
+			continue
+		}
+		timestamp := int(t.UnixNano() / 1000000)
```
collinvandyck: maybe rename this to `timestampMs` for clarity?

plugins/prometheus/prometheus.go
```diff
@@ -91,11 +97,20 @@ func stopPromServer(c *cli.Context) error {
...
 func promConnector(metrics []producers.MetricsMessage, c *cli.Context) error {
+	latestMetrics.Lock()
+	latestMetrics.metrics = metrics
+	latestMetrics.Unlock()
 	return nil
 }
 
 func serveMetrics(w http.ResponseWriter, r *http.Request) {
-	http.Error(w, "", http.StatusNoContent)
+	if len(latestMetrics.metrics) == 0 {
```
collinvandyck: Do we need to acquire the lock and create a copy here?  So something like

```
copy := []producers.MetricsMessage{}
latestMetrics.Lock()
copy = latestMetrics.metrics
latestMetrics.Unlock()
```

plugins/prometheus/prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func startPromServer(c *cli.Context) error {
+	addr := fmt.Sprintf(":%d", c.Int("prometheus-port"))
+	log.Infof("Starting Prometheus server on localhost%s", addr)
+
+	registerOnce.Do(func() {
+		// the server may start and stop, but the handler must only be
+		// registered once
+		http.HandleFunc("/metrics", serveMetrics)
```
collinvandyck: I think if you create an http.Server instance each time the prom server starts, you can register a handler for a path on that server instance instead of needing to set it globally here.

plugins/prometheus/prometheus_test.go
philipnrmn: Why would this flake?

collinvandyck: because something else could grab the port after you tested to see if it was available.  admittedly, not likely.

plugins/prometheus/prometheus.go
philipnrmn: The plugin starts and stops the server only once in its lifecycle, but the tests start it repeatedly. Unless you're strongly opposed, I'd prefer to leave it as it is.

philipnrmn:
@collinvandyck thanks for the review! I think I addressed all your feedback :)

plugins/prometheus/prometheus_test.go
philipnrmn: Gotcha. I had a look around, and other people needing a known free port use this strategy too (eg github.com/phayes/freeport). I think I'd prefer to risk very unlikely flakes in return for testing this all the way through. The alternative would be to do something like keeping a port free on our testing machines, I guess.

klueska:
What is left to get this pushed through so people can start using it?

plugins/prometheus/prometheus.go
```diff
@@ -0,0 +1,174 @@
...
+		if len(v) > 0 {
+			labels = append(labels, fmt.Sprintf("%s:%q", k, v))
+		}
+	}
+	for k, v := range tags {
+		labels = append(labels, fmt.Sprintf("%s:%q", k, v))
+	}
+
+	if len(labels) > 0 {
+		return "(" + strings.Join(labels, ",") + ")"
```
salekseev: I believe that per https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details the characters used to separate labels from the metric name are ```{}``` and not ```()```

salekseev:
Are there any plans to add protocol buffers output format in addition to text output as defined in https://prometheus.io/docs/instrumenting/exposition_formats/#format-variants-comparison that would  add significant efficiency as text format is meant mostly for human consumption?

collinvandyck:
:+1:, given that we address @salekseev's feedback.

plugins/prometheus/README.md
```diff
@@ -0,0 +1,71 @@
...
+Marathon by running
+
+`dcos marathon app add prometheus-server.json`
+
+You can also deploy it manually with the DC/OS UI.
+
+You will need to configure the Prometheus server to discover an endpoint on
+each of your nodes. You should do this by adding the IP address of each node
+to the `static_configs` array in the `scrape_configs` section of your
+`prometheus.yml` file. Remember to specify the port you're using too.
```
klueska: Can we add more info about how to inject a prometheus.yml file into the marathon app?

philipnrmn: Sure, I'll do that before merging.

plugins/prometheus/prometheus.go
philipnrmn: Thanks for catching this!

philipnrmn:
@salekseev re the protobuf output, we can certainly add that if there's appetite for it. A PR to enable this would be very welcome!

Text format is human readable, but it is not intended only for humans to read, so I'm happy to offer text format only in the first instance.

PR #129, opened 2017-11-20
Poll the metrics API via a system socket instead of adminrouter

philipnrmn:
Metrics plugins have until now polled the dcos-metrics API via their local adminrouter. 

This meant that the user had to specify the location of adminrouter (it's on different ports on agents and masters) and it also caused problems on enterprise DC/OS, where SSL certificates had to be specified. 

Removing this allowed a good deal of code cleanup to take place, and [plugin.go](../tree/master/plugins/plugin.go) is almost completely rewritten as a result. 

More concretely, this fixes plugins on master nodes, which were not able to authenticate previously.

- Adds [DCOS_OSS-1876](https://jira.mesosphere.com/browse/DCOS_OSS-1876)
- Fixes [DCOS_OSS-1828](https://jira.mesosphere.com/browse/DCOS_OSS-1828)

### DIFF
--- a/plugins/plugin.go
+++ b/plugins/plugin.go
@@ -174,7 +174,7 @@ func (p *Plugin) Metrics() ([]producers.MetricsMessage, error) {
 			URL:    &metricsURL,
 		}
 
-		metricMessage, err := makeMetricsRequest(p.Client, request)
+		metricMessage, err := makeMetricsRequest(p.Client, metricsURL.String())
 		if err != nil {
 			return metricsMessages, err
 		}
@@ -243,13 +243,16 @@ func (p *Plugin) setEndpoints() error {
 }
 
 /*** Helpers ***/
-func makeMetricsRequest(client *http.Client, request *http.Request) (producers.MetricsMessage, error) {
+
+// makeMetricsRequest polls the given url expecting to find a JSON-formatted
+// MetricsMessage, which it returns.
+func makeMetricsRequest(client *http.Client, url string) (producers.MetricsMessage, error) {
 	l := logrus.WithFields(logrus.Fields{"plugin": "http-helper"})
 
-	l.Infof("Making request to %+v", request.URL)
+	l.Infof("Making request to %+v", url)
 	mm := producers.MetricsMessage{}
 
-	resp, err := client.Do(request)
+	resp, err := client.Get(url)
 	if err != nil {
 		l.Errorf("Encountered error requesting data, %s", err.Error())
 		return mm, err
@@ -263,7 +266,7 @@ func makeMetricsRequest(client *http.Client, request *http.Request) (producers.M
 
 	// 204 No Content is not an error code; we handle it explicitly
 	if resp.StatusCode == http.StatusNoContent {
-		l.Warnf("Empty response received from endpoint: %+v", request.URL)
+		l.Warnf("Empty response received from endpoint: %+v", url)
 		return mm, nil
 	}
 
--- a/plugins/plugin.go
+++ b/plugins/plugin.go
@@ -21,7 +21,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"sync"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -164,23 +163,22 @@ func (p *Plugin) getNodeMetrics() (producers.MetricsMessage, error) {
 // endpoint for each container on the machine. It returns a slice of metrics
 // messages, one for each hit.
 func (p *Plugin) getContainerMetrics() ([]producers.MetricsMessage, error) {
+	var messages []producers.MetricsMessage
 	ids, err := p.getContainerList()
 	if err != nil {
 		return nil, errors.Wrap(err, "could not read list of containers")
 	}
-
-	var futures []<-chan producers.MetricsMessage
 	for _, id := range ids {
-		url1 := fmt.Sprintf("http://localhost/v0/containers/%s", id)
-		url2 := fmt.Sprintf("http://localhost/v0/containers/%s/app", id)
-		futures = append(futures, makeMetricsRequestFuture(p.Client, url1), makeMetricsRequestFuture(p.Client, url2))
+		containerMetrics, err := makeMetricsRequest(p.Client, fmt.Sprintf("http://localhost/v0/containers/%s", id))
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not retrieve metrics for container %s", id)
+		}
+		containerAppMetrics, err := makeMetricsRequest(p.Client, fmt.Sprintf("http://localhost/v0/containers/%s/app", id))
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not retrieve app metrics for container %s", id)
+		}
+		messages = append(messages, containerMetrics, containerAppMetrics)
 	}
-
-	var messages []producers.MetricsMessage
-	for future := range mergeFutures(futures) {
-		messages = append(messages, future)
-	}
-
 	return messages, nil
 }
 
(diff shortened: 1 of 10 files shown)


plugins/plugin.go
```diff
@@ -263,7 +266,7 @@ func makeMetricsRequest(client *http.Client, request *http.Request) (producers.M
 
 	// 204 No Content is not an error code; we handle it explicitly
 	if resp.StatusCode == http.StatusNoContent {
-		l.Warnf("Empty response received from endpoint: %+v", request.URL)
+		l.Warnf("Empty response received from endpoint: %+v", url)
```
collinvandyck: We can just use `%s` here for formatting since it's a string.  Same with line 252 above.

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
+// Metrics polls the local dcos-metrics API and returns a slice of
 // producers.MetricsMessage.
 func (p *Plugin) Metrics() ([]producers.MetricsMessage, error) {
+	var messages []producers.MetricsMessage
 
-	metricsMessages := []producers.MetricsMessage{}
+	// Fetch node metrics
+	nodeMetrics, err := p.getNodeMetrics()
+	if err != nil {
+		return messages, err
```
collinvandyck: I think it's more common to return `nil, err` in this situation unless there's a partial result.

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
+// getContainerMetrics polls the /containers/<id> and /containers/<id>/app
+// endpoint for each container on the machine. It returns a slice of metrics
+// messages, one for each hit.
+func (p *Plugin) getContainerMetrics() ([]producers.MetricsMessage, error) {
+	var messages []producers.MetricsMessage
+	ids, err := p.getContainerList()
+	if err != nil {
+		return messages, errors.Wrap(err, "could not read list of containers")
+	}
+	for _, id := range ids {
```
collinvandyck: Does it make sense to fetch the metrics for these IDs concurrently (with some sort of bound on the concurrency)?

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
 
-		metricMessage, err := makeMetricsRequest(p.Client, metricsURL.String())
+// getContainerMetrics polls the /containers/<id> and /containers/<id>/app
+// endpoint for each container on the machine. It returns a slice of metrics
+// messages, one for each hit.
+func (p *Plugin) getContainerMetrics() ([]producers.MetricsMessage, error) {
+	var messages []producers.MetricsMessage
+	ids, err := p.getContainerList()
+	if err != nil {
+		return messages, errors.Wrap(err, "could not read list of containers")
```
collinvandyck: I'd probably return `nil` for `messages` here as well as elsewhere in the code (I don't want to spam this comment everywhere).

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
+	}
+	for _, id := range ids {
+		containerMetrics, err := makeMetricsRequest(p.Client, fmt.Sprintf("http://localhost/v0/containers/%s", id))
```
collinvandyck: Alternatively, we could fetch these two metrics (container && app) here using two separate goroutines.

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
 		if err != nil {
-			return metricsMessages, err
+			return messages, errors.Wrapf(err, "could not retrieve app metrics for container %s", id)
 		}
+		messages = append(messages, containerMetrics, containerAppMetrics)
+	}
+	return messages, nil
+}
+
+// getContainerList polls the /containers endpoint and returns a slice of
```
collinvandyck: `polls` usually means you're going to hit it over and over, at least how I'm used to using the word.  Perhaps `queries` is better here?

plugins/plugin.go
```diff
@@ -151,39 +151,85 @@ func (p *Plugin) StartPlugin() error {
...
 
-		metricsMessages = append(metricsMessages, metricMessage)
-		p.Log.Infof("Received data from metrics service endpoint %s, success!", request.URL.Path)
+	resp, err := p.Client.Get("http://localhost/v0/containers")
+	if err != nil {
+		return ids, err
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
collinvandyck: You can replace some of this with:

```
err := json.NewDecoder(resp.Body).Decode(&ids)
return ids, errors.Wrap(err, "could not decode response body")
```

philipnrmn: I did not know about this nuance of errors.Wrap. Neat!

collinvandyck: I just found it out myself today!  Good stuff.

philipnrmn:
@collinvandyck I've addressed all your feedback and got a fanout working so that all requests are made in parallel.

plugins/plugin.go
```diff
@@ -199,6 +201,45 @@ func (p *Plugin) getContainerList() ([]string, error) {
...
+	}()
+	return out
+}
+
+// makeMetricsRequestFuture wraps makeMetricsRequest in a goroutine, returning
+// a channel on which the MetricsMessage will be returned.
+func makeMetricsRequestFuture(client *http.Client, url string) <-chan producers.MetricsMessage {
+	out := make(chan producers.MetricsMessage, 1)
+	go func() {
+		message, _ := makeMetricsRequest(client, url)
```
collinvandyck: Why are we ignoring the `err` return here?  Typically i would create a struct like

```
type messageAndErr struct {
  message producers.MetricsMessage
  err error
}

```

And then create a chan of these structs to return both the message and the error.  This way you do not have to ignore the error.

plugins/plugin.go
```diff
@@ -163,22 +164,23 @@ func (p *Plugin) getNodeMetrics() (producers.MetricsMessage, error) {
...
-			return nil, errors.Wrapf(err, "could not retrieve metrics for container %s", id)
-		}
-		containerAppMetrics, err := makeMetricsRequest(p.Client, fmt.Sprintf("http://localhost/v0/containers/%s/app", id))
-		if err != nil {
-			return nil, errors.Wrapf(err, "could not retrieve app metrics for container %s", id)
-		}
-		messages = append(messages, containerMetrics, containerAppMetrics)
+		url1 := fmt.Sprintf("http://localhost/v0/containers/%s", id)
+		url2 := fmt.Sprintf("http://localhost/v0/containers/%s/app", id)
+		futures = append(futures, makeMetricsRequestFuture(p.Client, url1), makeMetricsRequestFuture(p.Client, url2))
```
collinvandyck: A problem here is that you are creating an unbounded number of goroutines.  This can have resource exhaustion consequences at the client and possibly also the server due to resource exhaustion.  I'd pick a magic number and limit the concurrency to that number of goroutines.

philipnrmn:
Hi @collinvandyck, I reverted that last commit. Seeing as this is a time-sensitive PR, and the parallel requests are an optimisation that wasn't part of the intent of the original PR, maybe we could do it separately?

collinvandyck:
@philipnrmn hey yeah that makes sense. would it be worth it to log or instrument the time spent fetching all of the metrics so that we can later know if it's something worth optimizing?

philipnrmn:
@collinvandyck yes absolutely - that's a great idea.

philipnrmn:
Retest this please

PR #131, opened 2017-12-05
Fix permissions in the Prometheus plugin systemd configurations

philipnrmn:
Does what it says on the tin.

 - Resolves [DCOS_OSS-1942](https://jira.mesosphere.com/browse/DCOS_OSS-1942)


surdy:
👍 🚢

PR #133, opened 2017-12-14
Add blkio statistics to metrics

philipnrmn:
This commit adds blkio statistics to the `/containers/<container-id>` endpoint. 

If Mesos provides a blkio stats object (as shown in mesos.apache.org/documentation/latest/isolators/cgroups-blkio/) in the output of the container statistics endpoint, then metrics will offer blkio statistics for that container:

screen shot 2017-12-14 at 8 57 18 am

- Adds [DCOS_OSS-1489](https://jira.mesosphere.com/browse/DCOS_OSS-1489)

cc @bergerx


collectors/mesos/agent/agent_test.go
```diff
@@ -270,6 +270,37 @@ func TestTransform(t *testing.T) {
 			}
 			result := mac.metricsMessages()
 			So(len(result), ShouldEqual, 1) // one container message
+
+			expected_total_stats := map[string]uint64{
```
collinvandyck: I usually try to put different values into the fixture for each thing i want to test against.  this prevents a certain type of error where you may accidentally translate one value as a different value.

collectors/mesos/agent/agent_test.go
```diff
@@ -270,6 +270,37 @@ func TestTransform(t *testing.T) {
...
+			}
+			// Build map of device : names
+			actual_total_stats := map[string]uint64{}
+			for _, d := range result[0].Datapoints {
+				v, _ := d.Value.(uint64)
+					actual_total_stats[d.Name] = v
+					continue
+			}
+
+			for name, expected := range expected_total_stats {
```
collinvandyck: Do we want to check the value as well here (i.e. `expected`) ?

collinvandyck: ok, this was changed in a later commit.

philipnrmn:
Thanks very much for the review! I varied the test data as you suggested.

PR #136, opened 2017-12-20
Prometheus producer

philipnrmn:
This PR introducers a producer for Prometheus - a first-class Prometheus endpoint with no need to install third-party tools. 

By default the Prometheus endpoint is found on each node on port 9273, the same port as used as the default Prometheus port in telegraf. 

As we are using the prometheus_client library, the metrics endpoint supports both text and protobufs.


philipnrmn:
`http.ErrServerClosed` is not available in go 1.7. I updated the build server to run with go 1.8.

philipnrmn:
retest this please

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(
+		registry, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}))
+
+	server := &http.Server{
+		Addr:    net.JoinHostPort("localhost", strconv.Itoa(p.config.Port)),
+		Handler: mux,
+	}
+
+	go func() {
```
collinvandyck: The `http` producer actually blocks on `server.ListenAndServe()` and returns that error for the overall `Run() error` function.

It seems here that we are early exiting from the method as soon as we spin off a `ListenAndServe` in a new goroutine.  Do we want to block here instead?

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+			if err != http.ErrServerClosed {
+				promLog.Printf("E! Error creating prometheus metric endpoint, err: %s\n",
+					err.Error())
+			}
+		}
+	}()
 	return nil
 }
+
+// Describe passes all the available stat descriptions to Prometheus; we
```
collinvandyck: Why do we need to use a dummy metric to implement `Describe`? Some more doco here explaining that would be helpful.

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+}
+
+// Collect iterates over all the metrics available in the store, converting
+// them to prometheus.Metric and passing them into the prometheus producer
+// channel, where they will be served to consumers.
+func (p *promProducer) Collect(ch chan<- prometheus.Metric) {
+	for _, obj := range p.store.Objects() {
+		message, ok := obj.(producers.MetricsMessage)
+		if !ok {
+			promLog.Warnf("Unsupported message type %v", obj)
```
collinvandyck: Do we want to use `%T` here to describe the message type instead?

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+			}
+
+			name := sanitizeName(d.Name)
+			val, err := coerceToFloat(d.Value)
+			if err != nil {
+				promLog.Warnf("Bad datapoint value %q: %s", d.Value, err)
+				continue
+			}
+			desc := prometheus.NewDesc(name, "DC/OS Metrics Datapoint", tagKeys, nil)
+			metric, err := prometheus.NewConstMetric(desc, prometheus.GaugeValue, val, tagVals...)
```
collinvandyck: Are we setting the timestamp on the `metric` anywhere? if our store holds more than one time series for a particular metric this could cause overwrites if we're not doing this.

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+			if err != nil {
+				promLog.Warnf("Could not create Prometheus metric %s", name)
```
collinvandyck: We probably also want to log the error here.

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+				}
+			}
+		}
+	}
+}
+
+// sanitizeName returns a metric or label name which is safe for use
+// in prometheus output
+func sanitizeName(name string) string {
+	return strings.ToLower(illegalChars.ReplaceAllString(name, "_"))
```
collinvandyck: this is my least favorite 'feature' of prometheus

producers/prometheus/prometheus.go
```diff
@@ -15,17 +15,261 @@
...
+	case int64:
+		return float64(i), nil
+	case uint:
+		return float64(i), nil
+	case uint32:
+		return float64(i), nil
+	case uint64:
+		return float64(i), nil
+	default:
+		return math.NaN(), errors.New("value could not be coerced to float64")
```
collinvandyck: Maybe return type type of `unk` in the error message here.

producers/prometheus/prometheus_test.go
```diff
@@ -17,16 +17,88 @@
...
 		So(pi, ShouldHaveSameTypeAs, &promProducer{})
 		So(pc, ShouldHaveSameTypeAs, make(chan producers.MetricsMessage))
 	})
 }
+
+func TestRun(t *testing.T) {
+	Convey("Should create a Prometheus server on the configured port", t, func() {
+		// Ensure the environment is clean
+		if err := os.Unsetenv("LISTEN_PID"); err != nil {
+			panic(err)
```
collinvandyck: Is panicking here appropriate in a test?  In other words, do we want to use a test framework method to ensure that err is nil?

producers/prometheus/prometheus_test.go
```diff
@@ -17,16 +17,88 @@
...
+		time.Sleep(250 * time.Millisecond)
+
+		resp, err := http.Get("http://" + net.JoinHostPort("localhost", strconv.Itoa(port)) + "/metrics")
+		So(err, ShouldBeNil)
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		So(err, ShouldBeNil)
+
+		So(string(body), ShouldContainSubstring, "datapoint_one")
```
collinvandyck: Maybe leave a marker or TODO here that we should verify the response format a little more exactly.

producers/prometheus/prometheus_test.go
philipnrmn: I've been using panic when something has gone awry in test setup, but the code under test didn't fail (yet). Maybe it's better just to use assertions instead though 🤔

producers/prometheus/prometheus.go
philipnrmn: Our store can't hold two metrics with the same name and different timestamps - the older one will be replaced by the newer one. The lack of timestamps seemed odd to me too, but it's what Prometheus recommend in their exporter documentation:

> ..scrapes should be synchronous. Accordingly, you should not set timestamps on the metrics you expose, let Prometheus take care of that. If you think you need timestamps, then you probably need the Pushgateway instead.

(from https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling)

collinvandyck: aha, ok makes sense. thanks for the clarification 👍

producers/prometheus/prometheus.go
philipnrmn: saaaaame

philipnrmn:
@collinvandyck thanks very much for the very quick and tolerant review!

I addressed all your feedback.

PR #138, opened 2018-02-05
Prune NaN values from metrics to send to DataDog

philipnrmn:
Before this fix, NaN values received from containers were not removed
from the metrics which were to be sent to DataDog. When they were
encoded to JSON before being sent, the JSON encoder barfed and the
entire tranche of metrics was not sent.

Now, NaNs are selectively pruned and valid metrics are sent.

Fixes DCOS-20683


philipnrmn:
@crackcode123 ready for review :)

PR #139, opened 2018-03-01
Multi-platform builds

collinvandyck:
JIRA: https://jira.mesosphere.com/browse/DCOS-21365

The idea is to standardize some common make targets across our projects, and also make sure they can build on both Mac and Linux (using Docker if necessary).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM golang:1.9.3
+
+RUN go get -u github.com/jstemmer/go-junit-report
+RUN go get -u github.com/smartystreets/goconvey
+RUN go get -u golang.org/x/tools/cmd/cover
+RUN go get -u github.com/axw/gocov/...
+RUN go get -u github.com/AlekSi/gocov-xml
+RUN go get -u golang.org/x/tools/cmd/goimports
+RUN go get -u github.com/golang/lint/golint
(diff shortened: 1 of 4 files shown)


producers/http/listener_darwin.go
```diff
@@ -0,0 +1,28 @@
...
+	"os"
+)
+
+func getListener(unsetEnv bool) ([]net.Listener, error) {
+	return nil, nil
+}
+
+func getFiles(unsetEnv bool) []*os.File {
+	return nil
+}
```
philipnrmn: nice

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.9.3
+
+RUN go get -u github.com/jstemmer/go-junit-report
+RUN go get -u github.com/smartystreets/goconvey
+RUN go get -u golang.org/x/tools/cmd/cover
+RUN go get -u github.com/axw/gocov/...
+RUN go get -u github.com/AlekSi/gocov-xml
+RUN go get -u golang.org/x/tools/cmd/goimports
+RUN go get -u github.com/golang/lint/golint
```
bamarni: what about using go dep instead? so that both docker and non-docker users have a standard way to install deps.

collinvandyck: @bamarni we could do that, possibly in another PR.

PR #140, opened 2018-03-08
Add task ID and task name to Datadog metadata

philipnrmn:
This PR ensures that we always send the task name and task ID to DataDog alongside metrics.

The end result is being able to split metrics out by task name:
screen shot 2018-03-08 at 3 32 47 pm


philipnrmn:
@mrmarvin mind giving this a 👍 ? It was on your wishlist.

hectorj2f:
We'd like to have this in soon :) 🏭

PR #142, opened 2018-04-10
Ensure that Prometheus names cannot begin with numbers

philipnrmn:
Some frameworks prefix metrics with hexidecimal UUIDs, leading to numeric initials in metrics. This is illegal in Prometheus. This PR adds a test and a fix. 

 - fixes [DCOS_OSS-2360](https://jira.mesosphere.com/browse/DCOS_OSS-2360) - Metric names can begin with numbers in dcos-metrics Prometheus endpoint

PR #141, opened 2018-03-23
Overhaul the docs

philipnrmn:
The docs were inconsistent and out of date (for example the [architecture doc](https://github.com/dcos/dcos-metrics/blob/master/docs/ARCHITECTURE.md), last updated before DC/OS 1.9 was released). This overhaul updates them to make them much more approachable, and to focus on answering questions like 'how do I use this software' rather than 'what data transport did you decide to use internally'.

Diffs are hard to read, so I'd suggest browsing from the README at https://github.com/dcos/dcos-metrics/tree/docs

- fixes [DCOS-21663](https://jira.mesosphere.com/browse/DCOS-21663) Overhaul docs for DC/OS Metrics

PR #143, opened 2018-04-16
Ensure that names cannot begin with numbers in Prometheus plugin

philipnrmn:
This PR copies the changes from #142 into the Prometheus plugin to fix the same issue - it was possible for numbers to prefix Prometheus label names, which is illegal in Prometheus and caused such metrics to be ignored by the server.


philipnrmn:
@amitaekbote you were kind enough to review #141, would you mind doing the same here?

PR #146, opened 2018-04-25
Prometheus plugin: set the port used in the prometheus template and improve the README

hectorj2f:
The port used in the prometheus example points to the port `61091`. I suggest to change that port to help improving the user experience when installing and configuring this plugin.

Moreover, I also included some steps to know how to install/configure this plugin for older DC/OS versions


philipnrmn:
Thanks for the PR Hector, however I'm a little confused. There's some perfectly good, recently written documentation linked from the README on how to install the Prometheus plugin on older clusters here:

https://github.com/dcos/dcos-metrics/blob/master/docs/quickstart/prometheus-dcos19.md

Can't we just use that?

hectorj2f:
@philipnrmn I didn't see them before. They miss several steps such as the usage of prometheus.env, and force users to go and change the systemd unit. IMHO, I think the steps you defined for the stastd plugin are good, so I decided to follow the same pattern here. I was also confusing to see the steps for one plugin, and not for the other (so I also thought that applying the same pattern will reduce confusion).

PR #149, opened 2018-05-05
1.10 label app stats

philipnrmn:
Backports #148 to 1.10


philipnrmn:
NB CI is failing because we changed the build since 1.10, not because of test failures

PR #150, opened 2018-05-05
1.11 label app stats

philipnrmn:
Backports #148 to 1.11


philipnrmn:
NB CI is failing because we changed the build since 1.11, not because of test failures

PR #153, opened 2018-05-21
Modified Windows DCOS-Metrics to use detect_ip and clusterid file from the default path set by the dcos-go

soccerGB:
There are three separate commits included in this PR for addressing the following three issues

1. Made the dcos-metrics uses the detect_ip and clusterid files from the default paths
2. Ignored error from deleting non-exiting build directory
3. Removed the Linux only coreos/go-systemd/activation package from Windows unitests run

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -100,7 +100,7 @@ if ($args.Count -eq 0)
     exit -1
 }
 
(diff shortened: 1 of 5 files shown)


philipnrmn:
OK to test

PR #155, opened 2018-06-04
Remove static "servicePort" parameters from quickstart JSONs so that Marathon can automatically assign a random servicePort

ably77:


philipnrmn:
Thanks!

PR #159, opened 2018-06-20
Changed the use of `vim` into `vi`

edgarlanting:
Changed the use of `vim` into `vi` since `vim` isn't available on all of our supported distributions by default.


philipnrmn:
I squashed all three of your PRs into one over at #161 - I'll close them out now. Thanks for the contribution!

PR #164, opened 2018-07-18
Add debug route

philipnrmn:
This PR adds a /debug route to the v0 API, which shows all metrics currently in the store. 

- Partially fixes [DCOS-37454](https://jira.mesosphere.com/browse/DCOS-37454)

### DIFF
--- a/producers/http/handlers.go
+++ b/producers/http/handlers.go
@@ -25,6 +25,23 @@ import (
 	"github.com/gorilla/mux"
 )
 
+func debugHandler(p *producerImpl) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		allMetrics := p.store.Objects()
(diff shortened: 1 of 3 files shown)


producers/http/handlers.go
```diff
@@ -25,6 +25,23 @@ import (
 	"github.com/gorilla/mux"
 )
 
+func debugHandler(p *producerImpl) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		allMetrics := p.store.Objects()
+		if len(allMetrics) == 0 {
+			httpLog.Error("/v0/debug - no content in store.")
```
branden: Is there a reason the handlers in this file hardcode their paths in error message instead of getting them from `r.URL.Path`?

philipnrmn: they should definitely do that

PR #168, opened 2018-12-10
Fix missing containers in Prometheus exporter responses

branden:
https://jira.mesosphere.com/browse/DCOS_OSS-3863


branden:
@janisz I'll let you know when it's ready for you to take a look.

branden:
@philipnrmn @gracedo @janisz Ready for review!

branden:
I rewrote Prometheus producer tests to be more thorough, so I'll point out the test case that applies to the fix in this PR: https://github.com/dcos/dcos-metrics/pull/168/files#diff-2caadd9cf74f55ff0ae58fa5d34872e7R97

collectors/framework/framework_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/dcos/dcos-metrics/collectors"
 	mesosAgent "github.com/dcos/dcos-metrics/collectors/mesos/agent"
 	"github.com/dcos/dcos-metrics/producers"
-	"github.com/dcos/dcos-metrics/schema/metrics_schema"
+	metricsSchema "github.com/dcos/dcos-metrics/schema/metrics_schema"
```
janisz: why only this import is changed but not the code?

producers/prometheus/prometheus_test.go
```diff
@@ -37,58 +41,185 @@ func TestNew(t *testing.T) {
...
+								&dto.LabelPair{Name: proto.String("executor_id"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("framework_id"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("framework_name"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("framework_principal"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("framework_role"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("hello"), Value: proto.String("world")},
+								&dto.LabelPair{Name: proto.String("hostname"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("mesos_id"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("task_id"), Value: proto.String("")},
+								&dto.LabelPair{Name: proto.String("task_name"), Value: proto.String("")},
```
janisz: Do we need all these labels in this test? It's noise

producers/prometheus/prometheus_test.go
```diff
@@ -37,58 +41,185 @@ func TestNew(t *testing.T) {
...
+				panic(err)
+			}
 
-		resp, err := http.Get("http://" + net.JoinHostPort("localhost", strconv.Itoa(port)) + "/metrics")
-		So(err, ShouldBeNil)
-		defer resp.Body.Close()
+			// Start the server.
+			p, metricsChan := New(Config{Port: port})
+			go p.Run()
+			time.Sleep(1 * time.Second)
```
janisz: Do we really need to sleep here? What are we waiting for? Please replace it with a channel (if possible) or reduce this to a millisecond.

producers/prometheus/prometheus_test.go
```diff
@@ -37,58 +41,185 @@ func TestNew(t *testing.T) {
...
+			So(err, ShouldBeNil)
 
-		// TODO (philipnrmn): verify the output better than this
-		So(string(body), ShouldContainSubstring, "datapoint_one")
-		So(string(body), ShouldContainSubstring, "123.456")
+			// Provide metrics to the server from the test case's input.
+			for _, metricsMessage := range testCase.input {
+				metricsChan <- metricsMessage
+			}
+			time.Sleep(250 * time.Millisecond)
```
janisz: ditto

producers/prometheus/prometheus_test.go
branden: I put them here deliberately; I don't think they're noise. They're an explicit declaration of the tags we expect on metrics exposed by the Prometheus producer, which I think is appropriate for a test.

collectors/framework/framework_test.go
branden: goimports was complaining about some unaliased imports; see https://github.com/dcos/dcos-metrics/pull/168/commits/7dea6ee57735317052ce91650982b80f6e286b00

producers/prometheus/prometheus_test.go
branden: This is how the test already behaves. It's not relevant to the bug I'm trying to fix with this PR, and this repo is deprecated anyway, so I'd like to avoid diving down that rabbit hole unless it presents a practical problem.

producers/prometheus/prometheus_test.go
branden: https://github.com/dcos/dcos-metrics/pull/168#discussion_r241854118

PR #169, opened 2018-12-14
(1.11) Provide same set of label dimensions for metrics with the same name

branden:
This is a 1.11 backport of #168.

DC/OS PR: https://github.com/dcos/dcos/pull/3832


branden:
@gracedo @janisz @philipnrmn I think this is ready for review. The windows build is failing, but it looks like the cause is a missing powershell build script, which is unrelated to this PR. Is that check expected to fail on the 1.11 branch?

philipnrmn:
LGTM. For what it's worth, this is the only PR that will make it into DC/OS - the master branch isn't used anywhere. Obviously makes sense to update in both places.

PR #174, opened 2019-01-31
Sirupsen/logrus -> sirupsen/logrus

branden:
This prevents some conflicts that may occur when dcos-metrics is used as a dependency of a project that explicitly depends on sirupsen/logrus (lowercase).

This also syncs dcos/dcos-go, which `govendor sync` wouldn't leave alone.


philipnrmn:
the ghost of sirupsen strikes again 🤦‍♂️

